PR #1, opened 2021-01-25
Cli Migration from Freshli

dan-hein:
### Goal
The primary goal of these changes is to take the Cli components from Freshli and migrate them into this repo. 

### Details
To start, the Program.cs file and OutputFormatter.cs file have been migrated along with their git histories and tests into this repo and into new corresponding projects and solutions. To complement and solidify this change, on the cli-breakout branch, Freshli has been converted into a class library, and the migrated files have been removed.

The README has also been updated to reflect the proper build, run, and test steps and output.

### Next Steps
I believe this is ready to merge, but if there are any necessary adjustments, please let me know and I'd be more than happy to correct those issues! I believe that once this branch is ready and merged we can continue on the necessary improvements for the Cli.


Freshli.Cli.Test/Freshli.Cli.Test.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.7.1" />
+    <PackageReference Include="xunit" Version="2.4.1" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="coverlet.collector" Version="1.3.0">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
dfar-io, line 19: So for this, the changes I'm making for Code Coverage reporting via CodeClimate will require the use of the MSBuild variant of Coverlet: https://github.com/coverlet-coverage/coverlet#msbuild-integration-suffers-of-possible-known-issue

I'll be updating the Core project to use it, so we should update it here as well so we can get coverage across projects.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "freshli"]
+	banch = cli-breakout
```
dfar-io, line 2: So I think this actually works (it pulls down the correct commit regardless).. but should be `branch`

dfar-io (changes requested):
A few more changes, and thinking we should add a task to improve coverage *after* merging this in (no reason to hold up the overall process).

Code coverage of the CLI currently is:
* Line 42.44%
* Branch 50%
* Method 40%

Would like to see it get to around 80%.

.gitmodules
dan-hein: That is definitely an issue, thanks!

mrbiggred:
@dan-hein I removed all but @dfar-io as reviewers to prevent confusion on who was supposed to review this issue.  I want to avoid the problem where everyone assigned to a task thinks someone else is going to do it.

That said, if you or Dave want another reviewer feel free to reach out to that person.

dfar-io (changes requested):
I might be leaving this comment twice? Just the coverlet.msbuild change and this is good to merge.

dfar-io (approved):
:+1:

PR #5, opened 2021-02-15
Adds VSCode Dev Container support

dfar-io:
Allows for Visual Studio Code's Dev Containers to be used, and sets the path to allow for development using a Docker image.

The Dev Container created will also install the following plugins to assist with testing:
* .NET Test Explorer
* Coverage Gutters


dan-hein (approved):
The configurations look clean, and they work for me in VS Code!

Issue #7, opened 2021-02-26
Create automated build for the executable


mrbiggred:
**Edit:** Changed this case to create an automated build for the CLI.  Didn't notice one did not exist yet.  Leaving the below notes on GitVersion as using GitVersion will be part of the build process to determine the exe version.

Use [GitVersion](https://github.com/GitTools/GitVersion) to generate the version number for this project.  Actually the plan is to use the [GitHub Action](https://github.com/marketplace/actions/gittools).  GitVersion will be setup to support the Cactus Branching model (see below for details) which requires a setup like:

```
# GitVersion.yml
branches:
  master:
    mode: ContinuousDeployment
    tag: 'alpha'
    increment: Minor
    prevent-increment-of-merged-branch-version: false
    track-merge-target: true
  release:
    tag: 'beta'
    increment: Patch
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
```

The idea is that only `release-X.Y` branches are tagged something like `vX.Y.Z`.  Then this tag is merged back to `main` GitVersion will calculate the next version number as `vX.Y+1.0.0`.

When a feature branch is merged to `release-X.Y` for a bug fix only the patch number will be incremented and the major, minor, and patch numbers left as is.

GitVersion generates several different version number formats such as:

- AssemblySemVer: 0.8.0.0
- NuGetVersionV2: 0.8.0-alpha0001
- MajorMinorPatch: 0.8.0

GitVersion can set the version in C# Assembly Version files now that they have fixed this [issue](https://github.com/GitTools/actions/issues/125).

----

An example of the Cactus or Trident Branching model:

![image](https://user-images.githubusercontent.com/17273852/109237356-9f809e80-778e-11eb-92e6-65ae1480e584.png)

mrbiggred:
@mfcorgi I updated this issues title to clarify it's purpose.  This issue is for updating the version on the exe.  So when when exe is built you should be able to right click on it in Windows and see it has a version of vX.Y.Z.

Now that I look I don't see any GitHub workflows for this repo.  That means this case just became creating a basic build.  The build should compile do the usual items: compile, run tests (if any), along with versioning the executable.  You should be able to borrow from the Freshli Lib [dotnet-core.yml](https://github.com/corgibytes/freshli-lib/blob/main/.github/workflows/dotnet-core.yml) file.

mfcorgi:
[...]

PR #11, opened 2021-03-18
(#8) - Add release-notes.yml workflow to github actions and GitVersion.yml to allow automatic versioning

mfcorgi:
- Solution for #8  and #7
- Updated reference to Freshli.Lib that was broken after we changed the Nuget Package Name

### DIFF
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -69,10 +69,10 @@ jobs:
           dotnet-gitreleasemanager export --token ${{ secrets.GITHUB_TOKEN }} -o 'corgibytes' -r 'freshli-cli' -f 'CHANGELOG.md'
           cat CHANGELOG.md
 
-      - uses: stefanzweifel/git-auto-commit-action@v4
-        if: steps.commit.outputs.changes_detected == true
+      - uses: stefanzweifel/git-auto-commit-action@v4        
+        if: steps.auto-commit-action.outputs.changes_detected == 'true'
         name: Commit Change Log and if it Changed
         with:
-          commit_message: Committing auto generated change log.
+          commit_message: Committing auto generated change log.          
           file_pattern: CHANGELOG.md
 
--- a/.github/workflows/release-notes.yml
+++ b/.github/workflows/release-notes.yml
@@ -66,11 +66,11 @@ jobs:
       # is not included in the GitReleaseManager action yet.
       - name: Generate Change Log
         run: |
+          git config --global core.autocrlf input
           dotnet-gitreleasemanager export --token ${{ secrets.GITHUB_TOKEN }} -o 'corgibytes' -r 'freshli-cli' -f 'CHANGELOG.md'
           cat CHANGELOG.md
 
-      - uses: stefanzweifel/git-auto-commit-action@v4        
-        if: steps.auto-commit-action.outputs.changes_detected == 'true'
+      - uses: stefanzweifel/git-auto-commit-action@v4                
         name: Commit Change Log and if it Changed
         with:
           commit_message: Committing auto generated change log.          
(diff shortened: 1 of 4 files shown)


mrbiggred:
@mfcorgi this issue only fixes #8.  I updated #7 to make to clear that it is for creating the automated build to create the executable.  Sorry for the confusion.

mrbiggred (changes requested):
This actually only fixes #8 and not #7.  I updated #7 to clarify it is for creating the automated build creating the exe.  Sorry for the confusion.

That aside the PR looks good.  I have just one request to make sure errors are raised if the release not generation fails.

Finally I also created a [v0.2.0] tag.  This let me test that the release not generation script was working.  A working script should create/update a [v0.3.0](https://github.com/corgibytes/freshli-cli/releases/tag/untagged-a8b91d2377533332e5fd) GitHub draft release and generate a [CHANGELOG.md](https://github.com/corgibytes/freshli-cli/blob/freshli-cli-8/CHANGELOG.md).

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      # This will generate the change log for all the GitHub Releases, feature
+      # is not included in the GitReleaseManager action yet.
+      - name: Generate Change Log
+        run: |
+          dotnet-gitreleasemanager export --token ${{ secrets.GITHUB_TOKEN }} -o 'corgibytes' -r 'freshli-cli' -f 'CHANGELOG.md'
+          cat CHANGELOG.md
+
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        if: steps.commit.outputs.changes_detected == true
```
mrbiggred, line 73: Does this work?  It relies on a step called `commit` but I don't see it in the file.  I think the `steps.commit.outputs.changes_detected` will always be false.  I think this check is designed to be used after the auto commit as run.  For example:

```yml
# Notice the id value.
- uses: stefanzweifel/git-auto-commit-action@v4
  id: auto-commit
  if: steps.commit.outputs.changes_detected == true
  name: Commit Change Log and if it Changed
  with:
    commit_message: Committing auto generated change log.
    file_pattern: CHANGELOG.md

# Id value used in future step.
- name: "Run if changes have been detected"
  if: steps.auto-commit-action.outputs.changes_detected == 'true'
  run: echo "Changes!"

- name: "Run if no changes have been detected"
  if: steps.auto-commit-action.outputs.changes_detected == 'false'
  run: echo "No Changes!"
```

More details on the `changes_detected` output from the Git Auto Commit:

https://github.com/stefanzweifel/git-auto-commit-action#outputs

mrbiggred (changes requested):
I don't think the check you have for the auto commit action is working the way you expect.  Please double check and let me know if I'm mistaken.

We will also eventually have to incorporate the changes in corgibytes/freshli-lib#287.  Specifically the personal access token (PAT) change.  This will be required when [protected branches](https://docs.github.com/en/github/administering-a-repository/about-protected-branches) are turned on for Freshli CLI.  I'm waiting until all the protected branches issues are worked out in Freshli-Lib before turning them on in CLI.

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,77 @@
...
+          owner: 'corgibytes'
+          repository: 'freshli-cli'
+          milestone: 'v${{ steps.gitversion.outputs.majorMinorPatch }}' 
+
+      # This will generate the change log for all the GitHub Releases, feature
+      # is not included in the GitReleaseManager action yet.
+      - name: Generate Change Log
+        run: |          
+          dotnet-gitreleasemanager export --token ${{ secrets.GITHUB_TOKEN }} -o 'corgibytes' -r 'freshli-cli' -f 'CHANGELOG.md'
+          git add --renormalize CHANGELOG.md
```
mrbiggred, line 70: 👍

mrbiggred (approved):
Hopefully this kills the change log line ending issue once and for all.

Issue #9, opened 2021-02-26
Make the command line executable downloadable as a GitHub release asset


mrbiggred:
Compile and make the command line executable downloadable.  Not sure where the best place to put the download is?  For alpha releases my first thought would be GitHub either in the releases or possibly packages (if even possible).  We will need to discuss where to put beta and production versions of the executables.

mrbiggred:
@mfcorgi I was discussing with @mscottford some places to put the exe to download.  I think a good place to start is to put the exe as an GitHub release asset.  In the next month or so we can find another place to put the executable, such as on the Freshli.io website.

https://docs.github.com/en/github/administering-a-repository/about-releases

mrbiggred:
Closed by PR #17.

PR #12, opened 2021-03-23
(#7 and #9) - Add ci.yml workflow

mfcorgi:
- Consolidated dotnet-core.yml and release-notes.yml into ci.yml. (PR for #7 and #9)

This worfklow will work on pull_requests and pushes to main and releases branches. 

- It Will update the release draft only on push.
- It will always upload the alpha artifacts to the workflow run.

The idea is to create another workflow to transform the draft release into an actual itself. It could be when pushing a release branch into master.


.github/workflows/dotnet-core.yml
```diff
@@ -0,0 +1,55 @@
...
+    
+      - name: Display updated csproj File
+        run: cat Freshli.Cli/Freshli.Cli.csproj
+
+      - name: Build
+        run: dotnet build --configuration Release
+                
+      - name: Package
+        if: github.event_name == 'push'
+        run: dotnet pack -c Release
```
mrbiggred, line 51: The CLI does not produce a NuGet package.  The "Package" and "Publish Alpha Package to GitHub Packages" can be removed.

mrbiggred (changes requested):
The Command Line Interface (CLI) does not produce a NuGet package.  The steps that pack and publish the NuGet package should be removed from the build.

The result of this build should be an executable (exe) with the correct version number set.  Case #9 is to put the built exe somewhere after it is built.

mrbiggred (changes requested):
This looks the same as PR #17.  Should one of the PRs be discarded?

mfcorgi:
Closing this PR because the ci.yml file was already pushed in #17

PR #14, opened 2021-03-23
(#9) - Add executable assets

mfcorgi:


mrbiggred (commented):
I understand the problem you posted about in Slack.  The `dotnet-core.yml` builds the exe but the `release-notes.yml` creates the GitHub release.  How do we add the exe artifact to the GitHub Release if they are in sperate files.  Not sure but we can chat about it more in Slack and figure something out.

mfcorgi:
I am going to replace this PR with other one.

PR #16, opened 2021-03-23
(#15) - Add cla.yml workflow

mfcorgi:


.github/workflows/cla.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: "CLA Assistant"
+        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
+        uses: cla-assistant/github-action@v2.1.1-beta
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # the below token should have repo scope and must be manually added by you in the repository's secret
+          PERSONAL_ACCESS_TOKEN : ${{ secrets.CLA_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: 'CLA/signatures/v1.0/cla.json'
+          path-to-cla-document: 'https://github.com/corgibytes/freshli/blob/main/CLA/CLA-v1.0.md'
```
mrbiggred, line 21: We need to check with @mscottford if we we going to have one CLA for all of Freshli or one per repo?  If we have one per repo then the above path will need to point to Freshli-CLI.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,31 @@
...
+          branch: 'main'
+          allowlist: dependabot[bot]
+
+          #Configurable inputs
+          #remote-organization-name: enter the remote organization name where the signatures should be stored (Default is storing the signatures in the same repository)
+          #remote-repository-name:  enter the  remote repository name where the signatures should be stored (Default is storing the signatures in the same repository)
+          create-file-commit-message: 'Adds file for storing CLA Signatures'
+          signed-commit-message: 'Adds CLA signature for $contributorName (PR #$pullRequestNo)'
+          custom-notsigned-prcomment: 'Thank you for your submission to Freshli! Like many open-source projects, we ask that you agree to our [Contributor License Agreement](https://github.com/corgibytes/freshli/blob/main/CLA/CLA-v1.0.md) before we can accept your contribution. You can sign the CLA by posting a comment in this PR matching the text below:'
```
mrbiggred, line 30: Same as the above review comment about one CLA for all of Freshli.

mrbiggred (changes requested):
We need to check with @mscottford if there is only on CLA for all of Freshli or one per repo.

.github/workflows/cla.yml
mscottford: One per repo seems cleanest.

mscottford: Well, cleanest for us at least. Hopefully, it's not too annoying for others.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,31 @@
+name: "CLA Assistant"
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [opened,closed,synchronize]
+
+jobs:
+  CLAssistant:
+    runs-on: ubuntu-latest
```
mrbiggred, line 10: Should change this to be a specific ubuntu version such as Ubuntu 20.04.

PR #17, opened 2021-03-23
(#12) - Add codeclimate + codecoverage

mfcorgi:


.github/workflows/dotnet-core.yml
```diff
@@ -0,0 +1,65 @@
...
+          coverageLocations: |
+            ${{github.workspace}}/Freshli.Cli.Test/coverage.info:lcov
+
+      - name: Package
+        if: github.event_name == 'push'
+        run: dotnet pack -c Release
+                  
+      - name: Publish Alpha Package to Github Packages
+        if: github.event_name == 'push'
+        run: dotnet nuget push ./Freshli.Cli/bin/Release/*.nupkg -k ${{ secrets.GITHUB_TOKEN }} --skip-duplicate -n true -s https://nuget.pkg.github.com/corgibytes/index.json
```
mrbiggred, line 65: The above package creating and push can be removed.  There is no NuGet package for the command line.

.github/workflows/dotnet-core.yml
```diff
@@ -0,0 +1,65 @@
...
+                
+      - name: Test, Generate Code Coverage  
+        if: ${{ !env.ACT }}
+        uses: paambaati/codeclimate-action@v2.7.5
+        env:
+          CC_TEST_REPORTER_ID: ${{ secrets.TEST_REPORTER_ID }}
+        with:
+          coverageCommand: dotnet test /p:CollectCoverage=true /p:CoverletOutputFormat=lcov
+          coverageLocations: |
+            ${{github.workspace}}/Freshli.Cli.Test/coverage.info:lcov
```
mrbiggred, line 57: 👍

.codeclimate.yml
```diff
@@ -0,0 +1,3 @@
+exclude_patterns:
+  - "Freshli.Cli/obj/**"
+  - "Freshli.Cli.Test/obj/**"
```
mrbiggred, line 3: 👍

mrbiggred (changes requested):
This PR might need to be discarded and re-created depending on how we deal with #7 and #9.  Please also add the Code Climate badges to the Readme like in Freshli Lib:

https://github.com/corgibytes/freshli-lib/blob/main/README.md

mrbiggred (approved):
The PR looks fine to me.  I'm curious that the difference between this PR and PR #12 are?  It appears like the `ci.yml` files are the same.  Should one PR be deleted?

I do have a question about why the release is only updated on pushes and not PRs?  Is that to prevent the release from getting updated if a PR is not merged?

@dfar-io originally convinced me to have a separate release notes job in #corgibytes/freshli-lib#249.  Now you have convinced me to combine it back to one file.  Like you I can't think of a good way to upload the exe in a separate GitHub action.  Maybe we, or someone else, will come up with a better solution but for now one CI file is good.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,152 @@
...
+      - name: "[Post Build] - Add Assets to Release Draft"
+        if: ${{ github.event_name == 'push' }}
+        uses: gittools/actions/gitreleasemanager/addasset@v0.9.9
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          owner: 'corgibytes'
+          repository: 'freshli-cli'
+          milestone: 'v${{ steps.gitversion.outputs.majorMinorPatch }}'
+          tagName: 'v${{ steps.gitversion.outputs.majorMinorPatch }}'
+          assets: 'freshli-cli-${{ steps.gitversion.outputs.majorMinorPatch }}-win-x64.zip,freshli-cli-${{ steps.gitversion.outputs.majorMinorPatch }}-linux-x64.zip,freshli-cli-${{ steps.gitversion.outputs.majorMinorPatch }}-osx-x64.zip'
```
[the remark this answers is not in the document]
mfcorgi: It replaces an asset if already exists or upload a new asset. It will not delete assets with different names than the one being uploaded.

mfcorgi:
> The PR looks fine to me. I'm curious that the difference between this PR and PR #12 are? It appears like the `ci.yml` files are the same. Should one PR be deleted?
> 
> I do have a question about why the release is only updated on pushes and not PRs? Is that to prevent the release from getting updated if a PR is not merged?
> 
> @dfar-io originally convinced me to have a separate release notes job in #[corgibytes/freshli-lib#249](https://github.com/corgibytes/freshli-lib/pull/249). Now you have convinced me to combine it back to one file. Like you I can't think of a good way to upload the exe in a separate GitHub action. Maybe we, or someone else, will come up with a better solution but for now one CI file is good.

Yes, it is the same file. I was going to rebase once the other PR is approved. I added it here because this steps needs to be added in that file. We can close the other PR and merge only this and that will work.

Regarding a different files. Both previous files were applied on the same situation. (push or pr to master or release branche) I believe we need to keep one to upload draft artifacts, which is part of the CI workflow. For sure we need a different release workflow but to create a final release.

Regarding your other sentence, yes. you are correct, I do not update the draft release on MR because

1) we can have more than 1 PR for the same branch, which will override the artifact and we are never going to know which is the one attached  to the draft. 
2)  To prevent the release from getting updated if a PR is not merged.

If we want the executable for that particular workflow, we can go to the workflow artifacts.

PR #67, opened 2022-05-17
'cache prepare' command

### LINKED ISSUE #49: Add `cache prepare` command
## `cache prepare`

Used to ensure that the cache directory exists and that it contains a valid cache database instance. If an existing cache database is found, then the database will be checked to determine if migrations need to be run against it.

```
freshli [global options] cache prepare
```

**What this command does?**

Ensures that the cache directory exists and that the database that it contains is valid and up-to-date.

Global Otions

- `--cache-dir <path>`

CodeMouse92:
This implements the following features:

* `--cache-dir` global option
* `cache` command (cannot be directly executed)
* `cache prepare` subcommand, which idempotently creates or verifies the existence of a fully migrated SQLite3 database in the `--cache-dir`.
* Improves architecture to make it easier to add subcommands.

Closes #49.


mscottford:
I have read the CLA Document and I hereby sign the CLA

mscottford:
recheck

Corgibytes.Freshli.Cli.Test/Common/TestHelpers.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static void VerifyAlias<T> (string alias, ArgumentArity arity, bool allowMultipleArgumentsPerToken) where T: Command, new()
+    {
+        T command = new T();
+        Option option = command.Options.FirstOrDefault(x => x.Aliases.Contains(alias));
+        option.Should().NotBeNull();
+        option.AllowMultipleArgumentsPerToken.Should().Be(allowMultipleArgumentsPerToken);
+        option.Arity.Should().BeEquivalentTo(arity);
+    }
+}
```
mscottford, line 17: You didn't create this, so I think this little bit of clean-up can be pushed into a code clean-up issue.

Since we're using the `FluentAssertions` library, it would be cleanest to rely on [its support for adding new matchers](https://fluentassertions.com/extensibility/).

So this would end up looking like something like this: (I haven't tried to compile this.)

```csharp
public static class Command
{
    public static CommandOptionAssertions ShouldContainOptionAlias(this Command command, string alias)
    {
        return new CommandAssertions(command, alias);
    }
}

public class CommandOptionAssertions: ReferenceTypeAssertions<Option, CommandOptionAssertions>
{
    private string alias;
    private Command command;
    public CommandOptionAssertions(Command command, string alias): base(command)
    {
        this.command = command;
        this.alias = alias;
    }

    public AndConstraint<CommandAssertions> DoesAllowMultipleArgumentsPerToken()
    {
        Option option = command.Options.FirstOrDefault(x => x.Aliases.Contains(alias));
        Execute.Assertion
            .ForCondition(option != null)
            .FailWith("Alias with name {0} was not found", _ => alias)
            .Then
            .ForCondition(option.AllMultipleArgumentsPerToken)
            .FailWith("Expected to support multiple arguments per token")

        return new AndConstraint<CommandAssertions>(this);
    }

    public AndConstraint<CommandAssertions> DoesNotAllowMultipleArgumentsPerToken()
    {
        Option option = command.Options.FirstOrDefault(x => x.Aliases.Contains(alias));
        Execute.Assertion
            .ForCondition(option != null)
            .FailWith("Alias with name {0} was not found", _ => alias)
            .Then
            .ForCondition(!option.AllMultipleArgumentsPerToken)
            .FailWith("Expected to _not_ support multiple arguments per token")

        return new AndConstraint<CommandAssertions>(this);
    }

    public AndConstraint<CommandAssertions> HavingArity(int arity)
    {
        Option option = command.Options.FirstOrDefault(x => x.Aliases.Contains(alias));
        Execute.Assertion
            .ForCondition(option != null)
            .FailWith("Alias with name {0} was not found", _ => alias)
            .Then
            .BecauseOf(arity)
            .ForCondition(option.Arity == arity)
        return new AndConstraint<CommandAssertions>(this);
    }
}

```

And then I'm pretty sure it would be invoked like this:

```csharp
[Theory]
[InlineData("--output")]
[InlineData("-o")]
public void Verify_output_options_configuration(string alias)
{
    (new ScanCommand())
        .ShouldHaveOptionWithAlias("-o")
        .HavingArity(ArgumentArity.OneOrMore)
        .DoesAllowMultipleArgumentsPerToken();
}
```

Corgibytes.Freshli.Cli.Test/Extensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.CommandLine;
+using System.CommandLine.Parsing;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Corgibytes.Freshli.Cli.Test
+{
+    public static class Extensions
```
mscottford, line 11: The convention is to typically name these extension classes like this:

```suggestion
    public static class ParseResultExtensions
```

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
```diff
@@ -0,0 +1,37 @@
...
+        public CachePrepareCommandRunner(IServiceProvider serviceProvider, Runner runner)
+            : base(serviceProvider, runner)
+        {
+
+        }
+
+        public override int Run(CachePrepareCommandOptions options)
+        {
+            bool success = Cache.Prepare(options.CacheDir);
+            return (success == true ? 0 : 1);
```
mscottford, line 33: The `== true` isn't needed here and neither are the parentheses.

```suggestion
            return success ? 0 : 1;
```

Corgibytes.Freshli.Cli/CommandRunners/CommandRunner.cs
```diff
@@ -2,8 +2,8 @@
 using Corgibytes.Freshli.Lib;
 
 namespace Corgibytes.Freshli.Cli.CommandRunners
-{    
-    public abstract class CommandRunner<T> : ICommandRunner<T> where T : CommandOptions.CommandOptions 
+{
+    public abstract class CommandRunner<T> : ICommandRunner<T> where T : CommandOptions.CommandOptions
```
mscottford, line 6: Yay! More whitespace was removed.

Corgibytes.Freshli.Cli/CommandRunners/CommandRunner.cs
```diff
@@ -15,6 +15,6 @@ public CommandRunner(IServiceProvider serviceProvider, Runner runner)
         }
 
         public abstract int Run(T options);
-        
+
```
mscottford, line 18: And more still!

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    [Index(nameof(Key), IsUnique = true)]
+    public class CachedProperty
+    {
+        public int Id { get; set; }
+        public string Key { get; set; }
+        public string Value { get; set; }
+    }
+
+    class CacheContext : DbContext
```
mscottford, line 17: I'd like for us to stick to one class per file where the file name matches the class name.

features/cache.feature
```diff
@@ -0,0 +1,7 @@
+Feature: Cache
+    Scenario: Prepare
+        Given a directory named "~/.freshli" does not exist
+        When I run `Corgibytes.Freshli.Cli cache prepare`
```
mscottford, line 4: Let's add another scenario that changes the default cache folder location.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        public static readonly string cacheDirEnvVariable = "FRESHLI_CACHE_DIR";
```
mscottford, line 19: I like having the env variable as an option here. We should add a test for it, though. And we should clearly define whether or not the env variable overrides the command line parameter or vice-versa. I'm not sure what the convention should be there.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -0,0 +1,84 @@
...
+                string cacheDir = Environment.GetEnvironmentVariable(CacheContext.cacheDirEnvVariable);
+                return cacheDir.IsNullOrEmpty() ? CacheContext.cacheDirDefault.ToString() : cacheDir;
+            }
+        }
+
+        public DbSet<CachedProperty> CachedProperties { get; set; }
+
+        public CacheContext()
+        {
+            DbPath = System.IO.Path.Join(CacheDir, cacheDbName);
```
mscottford, line 37: Looks like this isn't utilizing the global option that is provided on the command line. Am I missing something?

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -0,0 +1,84 @@
...
+        public static readonly string cacheDirEnvVariable = "FRESHLI_CACHE_DIR";
+        public static readonly DirectoryInfo cacheDirDefault = new DirectoryInfo(Environment.GetEnvironmentVariable("HOME") + "/.freshli");
+        public static readonly string cacheDbName = "freshli.db";
+        public string DbPath { get; }
+
+        public static string CacheDir
+        {
+            get
+            {
+                string cacheDir = Environment.GetEnvironmentVariable(CacheContext.cacheDirEnvVariable);
```
mscottford, line 28: I'm pretty sure that `CacheContext` can be dropped here:

```suggestion
                string cacheDir = Environment.GetEnvironmentVariable(cacheDirEnvVariable);
```

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -0,0 +1,84 @@
...
+        public int Id { get; set; }
+        public string Key { get; set; }
+        public string Value { get; set; }
+    }
+
+    class CacheContext : DbContext
+    {
+        public static readonly string cacheDirEnvVariable = "FRESHLI_CACHE_DIR";
+        public static readonly DirectoryInfo cacheDirDefault = new DirectoryInfo(Environment.GetEnvironmentVariable("HOME") + "/.freshli");
+        public static readonly string cacheDbName = "freshli.db";
```
mscottford, line 21: In C#, the naming convention is for public fields, properties, methods, etc... to be `PascalCased`. Also, const is a better fit than `readonly` in when the value is a constant

```suggestion
        public static const string CacheDirEnvVariable = "FRESHLI_CACHE_DIR";
        public static readonly DirectoryInfo CacheDirDefault = new DirectoryInfo(Environment.GetEnvironmentVariable("HOME") + "/.freshli");
        public static const string CacheDbName = "freshli.db";
```

mscottford (changes requested):
Overall this looks really good, especially when you consider that you're not used to working with C#. You've picked up a ton of stuff to pull this PR together. :tada:

I do have some suggestions here and there. I _could_ have done many of them myself, but I wanted to take the opportunity to share some C# knowledge where it makes sense.

.codeclimate.yml
```diff
@@ -1,3 +1,5 @@
 exclude_patterns:
   - "Corgibytes.Freshli.Cli/obj/**"
-  - "Corgibytes.Freshli.Cli.Test/obj/**"
\ No newline at end of file
+  - "Corgibytes.Freshli.Cli.Test/obj/**"
+  - "Corgibytes.Freshli.Cli.Test/**"
```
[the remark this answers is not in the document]
CodeMouse92: The trouble is, it's giving us false alarms about duplicating the mandatory headings for tests. ;)

Corgibytes.Freshli.Cli/Functionality/Cache.cs
CodeMouse92: It IS using it, but I store the overridden path on an environment variable, so external calls to 'dotnet ef' are not broken. See line 64 for where I set that.

Corgibytes.Freshli.Cli.Test/CommandOptions/ScanCommandOptionsTest.cs
```diff
@@ -53,7 +55,7 @@ public void Send_Args_ReturnsScanOptions(string[] args, string expectedPath, For
             //Veryfy there aren't extra output stragetegies added
             outputStrategyTypes.Should().BeSubsetOf(expectedOutput);
         }
-
+
```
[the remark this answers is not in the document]
CodeMouse92: I'm trying to figure out how this got introduced, as I didn't change this.

Corgibytes.Freshli.Cli.Test/Common/TestHelpers.cs
CodeMouse92: Let's definitely make a separate task for this and move the notes there. I'm wary of trying to implement this right now.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
CodeMouse92: I've added a comment to clarify: this is principally used internally to store and retrieve the user-provided `--cache-dir` argument from the database. However, I'll adjust the code so if the argument is NOT provided, and the environment variable is, it will use the environment variable's value.

mscottford:
I'm confused about why this pull request is now reporting conficts.

Corgibytes.Freshli.Cli.Test/Common/TestHelpers.cs
mscottford: Sounds good. Issue #68 has been created to track this.

mscottford (approved):
Yay! First command in the bag!

PR #81, opened 2022-05-19
'cache destroy' command

### LINKED ISSUE #50: Add `cache destroy` command

CodeMouse92:
This implements the following features:

* `cache destroy` subcommand, which idempotently deletes the cache specified by `--cache-dir` (or the default thereof), without deleting non-Freshli-cache files.
* `--force` option on `cache destroy` for bypassing confirmation prompt.

Closes #50 .


Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
```diff
@@ -33,5 +34,37 @@ public override int Run(CachePrepareCommandOptions options)
...
+            : base(serviceProvider, runner)
+        {
+
+        }
+
+        public override int Run(CacheDestroyCommandOptions options)
+        {
+            // Skip prompt if the --force flag is passed
+            if (options.Force)
+                goto Destroy;
```
mscottford, line 50: Please use `{` and `}` explicitly. And rewrite this logic so that it does not rely on `goto`.

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
```diff
@@ -33,5 +34,37 @@ public override int Run(CachePrepareCommandOptions options)
...
+
+            // Prompt the user whether they want to destroy the cache
+            Console.Out.Write($"Do you want to destroy the Freshli cache at {options.CacheDir.ToString()}? [y/N] ");
+            string choice = Console.In.ReadLine();
+            string[] choicesToProceed = {"y", "Y"};
```
mscottford, line 55: The `var` keyword can be use here.

```suggestion
            var choice = Console.In.ReadLine();
            var choicesToProceed = {"y", "Y"};
```

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
```diff
@@ -33,5 +34,37 @@ public override int Run(CachePrepareCommandOptions options)
...
+            {
+                Console.Out.WriteLine("Operation aborted. Cache not destroyed.");
+                return 0;
+            }
+
+            // Destroy the cache
+            Destroy:
+            Console.Out.WriteLine("Destroying cache...");
+            bool success = Cache.Destroy(options.CacheDir);
+            return (success == true ? 0 : 1);
```
mscottford, line 67: The parentheses and `== true` can be removed here.

```suggestion
            return success ? 0 : 1;
```

Also, since this is the second place that we're translating a boolean to an exit code, I think it makes sense to create an extension method on `System.Boolean` that does just that. Something like this.

```csharp
public static class BooleanExtensions
{
    public static int ToExitCode(this bool value)
    {
        return value ? 0 : 1;
    }
}
```

And then the code in this location would change to:

```csharp
return Cache.Destroy(options.CacheDir).ToExitCode();
```

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -41,5 +47,40 @@ public static bool Prepare(DirectoryInfo cacheDir)
 
             return true;
         }
+
+        public static bool Destroy(DirectoryInfo cacheDir)
```
mscottford, line 51: This can be simplified significantly by just deleting the directory. We're going to have other commands adding lots of other files in here. I'm not sure it makes sense to keep track of every file that gets created.

features/cache.feature
```diff
@@ -12,3 +12,18 @@ Feature: Cache
...
+        Then the directory named "~/.freshli" should not exist
+
+    Scenario: Destroy on unclean folder
+        Given a directory named "~/.freshli" exists
+        And a blank file named "~/.freshli/freshli.db" exists
+        And a blank file named "~/.freshli/nonsense.txt" exists
+        When I run `freshli cache destroy --force`
+        Then the directory named "~/.freshli" should exist
+        And the file named "~/.freshli/nonsense.txt" should exist
+        And the file named "~/.freshli/freshli.db" should not exist
```
mscottford, line 29: I think we can get rid of this scenario.

features/cache.feature
```diff
@@ -12,3 +12,18 @@ Feature: Cache
         Then the directory named "~/somewhere_else" should exist
         And a file named "~/somewhere_else/freshli.db" should exist
         And we can open a SQLite connection to "~/somewhere_else/freshli.db"
+
+    Scenario: Destroy
```
mscottford, line 16: Let's add a destroy scenario that uses an alternative cache directory location.

features/step_definitions/database.rb
```diff
@@ -1,6 +1,30 @@
+require 'fileutils'
 require 'aruba/cucumber'
 require 'sqlite3'
 
+Given('that I ran {string}') do |command|
+    system command
+end
```
mscottford, line 7: I missed this before. But the convention in ruby is to use 2 spaces for indentation.

```suggestion
Given('that I ran {string}') do |command|
  system command
end
```

mscottford (changes requested):
Looks good. I think it can be made much simpler though. We don't need to keep an accounting of files that are stored in the cache directory.

I'm also going to create an issue to track adding ruby-specific rules to the `.editorconfig` file.

There are some other style choices that I noticed in here that I'd like to discourage the use of. I'll create an issue for adding some automation to help with remembering these:
1. Always use braces for control logic for `if`, `for`, `while`, etc...
```csharp
// avoid
if (expression)
    statement;

// embrace
if (expression)
{
    statement;
}
```
2. Avoid the use of `goto`.
3. Use `var` liberally.
4. Avoid the [query syntaxt](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/linq/) for linq expressions. Use the expanded form instead, usually defined in the [`System.Linq` namespace](https://docs.microsoft.com/en-us/dotnet/api/system.linq).
```csharp
//avoid
var values = from collection where value == true;

// embrace
var values = collection.Where(v => v == true);
```

Corgibytes.Freshli.Cli/Functionality/Cache.cs
CodeMouse92: I've taken extensive trouble on the design because of some rather serious consequences of "just delete the directory". I'm going to be stubborn about this one, because it's a matter of preventing system-fatal mistakes.

https://github.com/corgibytes/freshli-cli/issues/50#issuecomment-1129255629

features/cache.feature
CodeMouse92: No, because of the design mentioned in my response above.

features/cache.feature
CodeMouse92: I had tried, but I was having an unbelievable amount of trouble with the paths. After an hour, I gave up.

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
CodeMouse92: Is there a particular reason why you want to use `var`? Preference? Refactoring?

Personal position, I find explicit types to be more readable unless the type is appearing explicitly somewhere else in the line. In this case, I think `var` reduces readability, or more specifically, increases cognitive load while reading.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
mscottford: Hmm... the home directory is an interesting scenario to defend against. in that case, if `freshli.db` is present, I say we delete everything. Otherwise, we refuse to do anything. The presence of `freshli.db` should be sufficient for determining that it's an app created directory.

features/cache.feature
mscottford: I understand that it can be frustrating. But it still needs to be done.

mscottford: If you want a hand with it, please let me know.

features/cache.feature
mscottford: Roger. In that case, I think the logic will need to change a little bit. And we probably need to add a few additional scenarios.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
CodeMouse92: It shouldn't be sufficient, because the person could just as easily type...

```
freshli cache prepare --cache-dir /home/myself
```

...and then later...

```
freshli cache destroy --cache-dir /home/myself
```

I see two ways to proceed:

1. My approach.
2. Not allowing `cache prepare` in a non-empty directory unless it already contains `freshli.db`

Which would you prefer?

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
mscottford: It's perhaps more of a personal preference after working with languages with more flexible type systems. I like how readable the resulting code is. And if I'm curious about the exact type, there are many different ways to go about getting that information. 

Essentially, I'm a fan of using the `var` keyword everywhere that it can be used. If the compiler is not able to figure out the type, then it lets you know.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
mscottford: In that case, let's go with option 2.

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
mscottford: Thanks for making the change.

Corgibytes.Freshli.Cli/CommandRunners/CacheCommandRunner.cs
CodeMouse92: Noted. With all due respect, since I'm part of the team that will have to keep maintaining this, I really don't want to change this on that basis. I spend more time in Python than anything, and deeply love dynamic typing and duck-typing, but C# (like C++, which has `auto`) isn't that, and types can lead to a lot of surprises. If this comes down to preferences, the cognitive load is lower at least on me (and I know it is on many others) if `var` (like `auto`) is reserved for cases where the type is either explicit elsewhere in the line, or is particularly convoluted.

CodeMouse92:
I will address the rest of the comments tomorrow. Some style changes will appear in the next PR, although I'm trying to address as many in the cache-related code files here.

features/cache.feature
CodeMouse92: Added

CodeMouse92:
I do not think I can satisfy codeclimate on this point, because C# won't actually let me refactor these two constructors in the way it thinks I can. Open to suggestions.

mscottford:
> I do not think I can satisfy codeclimate on this point, because C# won't actually let me refactor these two constructors in the way it thinks I can. Open to suggestions.

I think that commit 7ff560b is a decent way to remove the duplication.

features/step_definitions/database.rb
```diff
@@ -1,11 +1,35 @@
...
-        expect(r[0]).to match(["ok"])
+Given('that I ran {string}') do |command|
+  system command
+end
+
+Given('a blank file named {string} exists') do |filename|
+  Aruba.configure do |config|
+    if filename.include? "~"
+      filename["~"] = config.home_directory
+    end
```
mscottford, line 13: The logic for expanding the `~` into the home directory is repeated a few times in this file. I think it makes sense to add a small method to do this.

```ruby
def expand_home_directory(filename, home_directory)
  if filename.include? "~"
    filename['~'] = home_directory
  end
  filename
end
```

And then this call site would change to:

```suggestion
    filename = expand_home_directory(filename, config.home_directory)
```

Corgibytes.Freshli.Cli/Commands/CacheCommand.cs
```diff
@@ -1,35 +1,39 @@
...
     {
-        public CacheCommand() : base("cache", "Manages the local cache database")
+        public CacheCommand()
+            : base("cache", "Manages the local cache database")
         {
             CachePrepareCommand prepare = new();
             AddCommand(prepare);
+
+            CacheDestroyCommand destroy = new();
+            AddCommand(destroy);
```
mscottford, line 15: These variables can be inlined:

```suggestion
            AddCommand(new CachePrepareCommand());
            AddCommand(new CacheDestroyCommand());
```

Corgibytes.Freshli.Cli/Commands/RunnableCommand.cs
```diff
@@ -0,0 +1,25 @@
...
+using Corgibytes.Freshli.Cli.CommandRunners;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Corgibytes.Freshli.Cli.Commands;
+
+#nullable enable
+public abstract class RunnableCommand<T> : Command where T: CommandOptions.CommandOptions
+{
+    protected RunnableCommand(string name, string? description) : base(name, description)
```
mscottford, line 13: I don't see `description` being set to null in this project so far. And if it was, I don't think it would look awesome to see `null` passed in as a parameter.

I think it makes sense to either drop `?` from `string?` for `description` or to provide a default value of `null` so that it can be optionally provided.

So either:

```suggestion
    protected RunnableCommand(string name, string description) : base(name, description)
```

or:

```suggestion
    protected RunnableCommand(string name, string? description = null) : base(name, description)
```

Corgibytes.Freshli.Cli/Commands/RunnableCommand.cs
```diff
@@ -0,0 +1,25 @@
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.CommandLine.NamingConventionBinder;
+using Corgibytes.Freshli.Cli.CommandRunners;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Corgibytes.Freshli.Cli.Commands;
+
+#nullable enable
```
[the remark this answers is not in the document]
mscottford: If you agree with making it project-wide, then please create an issue for it.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -1,32 +1,68 @@
...
 using System.Linq;
+using System.Runtime.Serialization;
 using Microsoft.EntityFrameworkCore;
 
 namespace Corgibytes.Freshli.Cli.Functionality
 {
+    [Serializable]
+    public class CacheException : Exception
+    {
+        public bool IsWarning { get; init; }
```
mscottford, line 13: I didn't have a chance to chat with you about this approach before I committed it. I'm curious what you think about it.

mscottford (changes requested):
Thanks for this. I think this is turning out nicely. And thanks for implementing the compromise for handling the `--cache-dir` option.

I made a bunch of small-ish suggestions, and I also noticed that the aruba/cucumber tests are focused on the `--force` option. I think it's important to also add scenarios that don't use `--force` so that we can confirm that the correct thing happens in those cases, specifically directories not getting deleted and warning/error messages getting written out to the console.

I also tried to point out the things that I liked, so that the review doesn't come across as overly negative.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
```diff
@@ -1,32 +1,68 @@
...
 
             // Create the directory if it doesn't already exist
-            if (cacheDir.Exists == false)
+            if (!cacheDir.Exists)
             {
                 cacheDir.Create();
             }
+            else if (!ValidateDirIsCache(cacheDir))
+            {
+                throw new CacheException($"We cannot use an existing non-empty directory as a cache directory.");
```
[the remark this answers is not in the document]
CodeMouse92: Do we have a way to route logging messages to the context provided to the runner? That has to be taken into account.

Corgibytes.Freshli.Cli/Commands/RunnableCommand.cs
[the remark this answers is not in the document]
CodeMouse92: I'd considered it, and even tried it briefly, but it would require a fair bit of refactoring. I'll create an issue to spike it.

Corgibytes.Freshli.Cli/Functionality/Cache.cs
CodeMouse92: Frankly, I think it detracts from readability, but no sense taking additional time to fix it now.

CodeMouse92:
Handed off; revise and merge at will.

mscottford (approved):
This looks good. Thanks!

PR #88, opened 2022-06-02
added version test command

donabp:
created a new version.feature file within features folder and added a scenario for freshli --version command


donabp:
recheck

mscottford (approved):
Looks good. This was worked on during a pairing session.

Issue #61, opened 2022-05-02
Add `agents detect` command


mscottford:
## `agents detect`

Used to detect all of the language agents that are available for use.

```
freshli agents detect
```

**What this command does?**

Language agents are detected by searching the `$PATH` for a program name that starts with `freshli-agent-`. It is assumed that anything after this prefix is the language name, and would be used whenever referring to the agent that directly.

This command outputs the detected language name and the path to the language agent binary in a tabular format. One line of output is created for each language agent that is detected.

mscottford:
To help with testing/building this command, there's a skeleton for the [Freshli Java Agent](https://github.com/corgibytes/freshli-agent-java) that can be installed in your path. You can also create a simple shell script that is named `freshli-agent-something`. A symbolic link could work, too.

PR #96, opened 2022-06-10
Added agents detect command issue #61

donabp:
Creating a pull request and submitting it for review for issue https://github.com/corgibytes/freshli-cli/issues/61 .


Corgibytes.Freshli.Cli/Commands/AgentsCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+                     filesResults = Directory.GetFiles(path);
+                 }
+                foreach(string file in filesResults){
+                    if (Path.GetFileName(file).Contains("freshli-agent-")) {
+                        agents.Add(file);
+                    }
+                }
+            }
+            
+            agents = agents.Distinct().ToList();
```
mscottford, line 59: This logic should be moved into a class named something like `AgentDetector` or something to that effect. That way we can more easily write unit tests against this functionality.

Corgibytes.Freshli.Cli/Commands/AgentsCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+        private int Run(IHost host, InvocationContext context, AgentsDetectCommandOptions options)
+        {
+            string [] paths =System.Environment.GetEnvironmentVariable("PATH").Split(":");
+            
+            List<string> agents = new List<string>();
+            string[] filesResults = null;
+
+            foreach (string path in paths) {
+                
+                 if(path.Contains("~/")){
```
mscottford, line 44: It's going to be best to store `~/` in a constant. That's something that could go in the `Environment` class. Perhaps calling it `Environment.HomeDirectoryPlaceholder`.

Corgibytes.Freshli.Cli/Commands/AgentsCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        public AgentsDetectCommand() : base("detect",
+            "Outputs the detected language name and the path to the language agent binary in a tabular format")
+        {
+            Handler = CommandHandler.Create<IHost, InvocationContext, AgentsDetectCommandOptions>(Run);
+        }
+
+        private int Run(IHost host, InvocationContext context, AgentsDetectCommandOptions options)
+        {
+            string [] paths =System.Environment.GetEnvironmentVariable("PATH").Split(":");
```
mscottford, line 37: Instead of using `:` here, it's best to use `Path.PathSeparator`.

mscottford (changes requested):
Yay! Your first pull request on the project! This is great!

I've made a _ton_ of comments here. Let's work on them together. Please don't read these and think that you've done a bad job. 

It's clear that the functionality works. You understood the problem and you implemented it. That's great! 

I just want us to make some adjustments to how the implementation is structured.

Corgibytes.Freshli.Cli/Commands/AgentsCommand.cs
donabp: are you saying I should create a new class called Environment and have Environment.HomeDirectoryPlaceholder as a field, or should I find an object with this field?

Corgibytes.Freshli.Cli/Commands/AgentsCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+                foreach(string file in filesResults){
+                    if (Path.GetFileName(file).Contains("freshli-agent-")) {
+                        agents.Add(file);
+                    }
+                }
+            }
+            
+            agents = agents.Distinct().ToList();
+            foreach(string agent in agents){
+                Console.WriteLine("The following agent was detected: " + agent );
```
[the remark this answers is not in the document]
donabp: did you want me just to add whitespace here or is there a method that I should be calling? im going to search through the codebase for some other examples of such output

edwinkortman (approved):
Approved as an ensemble with @donabp, @mscottford, and @CodeMouse92 .

PR #99, opened 2022-06-14
Clarify what the preferred way of setting up the development environment is

edwinkortman:
This PR originates from some struggles I had with setting up the environment. I saw the devcontainer environment, but I figured both methods described would work fine. Alas! I had some trouble setting up the first, so I spent longer than I wanted. To not have the next person make the same mistake as I did, I added some text to the README to clarify that the preferred way of setting up the development environment is the devcontainer. Also, if they run into the same error, how to solve it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ A tool for displaying historical metrics about a project's dependencies.  Run th
 
 ## Installing and Running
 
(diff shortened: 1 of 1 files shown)


edwinkortman:
I have read the CLA Document and I hereby sign the CLA

edwinkortman:
recheck

mscottford (approved):
Nice!

PR #110, opened 2022-06-15
Resolve Lint Errors

CodeMouse92:
Recreates most of the refactoring work originally done in #86. Adding additional linters should be done in a separate PR.


mscottford (changes requested):
I like these changes. However, they are not being flagged by the linters that are being run by `bin/lint.rb` as it is currently configured. If these are being reported by JetBrains Rider, then I’d like for us to add the [ReSharper command line tools](https://www.jetbrains.com/help/resharper/ReSharper_Command_Line_Tools.html) to `bin/lint.rb` and `bin/format.rb` while updating the `.editorconfig`. That will help us enforce these kinds of changes project wide for now and into the future, since `bin/lint.rb` is run as part of our CI process.

CodeMouse92:
Some of these were reported by SonarLint, which you already said you didn't want to adopt. I still think the code smells were worth changing, if only this once.

CodeMouse92:
@edwinkortman Tomorrow, we'll have to suss out why GH Actions is having a tantrum.

edwinkortman (approved):

PR #109, opened 2022-06-15
Git Clone Command (take 2)

CodeMouse92:
Resolves #51

PR #116, opened 2022-06-30
Checkout history

### LINKED ISSUE #54: Add `git checkout-history` command
## `git checkout-history`

Used to checkout a specific historical point for a given repository.

```
freshli [global options] git checkout-history [command options] <repository-id> <sha>
```

**What this command does?**

Delegates to the `git archive` command to create an archive of the specified sha, and then extract that archive into the `$CACHE_ROOT/histories/<repository id>/<sha>` directory.

Global Options

- `--cache-dir <path>`

Command Options

- `--git-path`
    - Path to the `git` binary
    - default value: `git`

edwinkortman:
Fixes #54

This PR adds the `git checkout-command` and this was partly paired with @donabp and @CodeMouse92. There's a small refactor of the Git bits added in https://github.com/corgibytes/freshli-cli/pull/109 as well as we came to the conclusion that this would be better when we are able to unit test it. :-) Otherwise it does what it says on the tin.

## `git checkout-history`

Used to checkout a specific historical point for a given repository.

```
freshli [global options] git checkout-history [command options] <repository-id> <sha>
```

**What this command does?**

Delegates to the `git archive` command to create an archive of the specified sha, and then extract that archive into the `$CACHE_ROOT/histories/<repository id>/<sha>` directory.

Global Options

- `--cache-dir <path>`

Command Options

- `--git-path`
    - Path to the `git` binary
    - default value: `git`


CodeMouse92:
I'm going to pick on commit messages for a moment. Most of these are reasonable (given that I hadn't shared about conventional commits etc yet), but could you possibly change the one that is simply `wiring` to be a bit more descriptive? `git rebase -i HEAD~7` should let you select that one for rewriting the message; this is one case where rewriting history should be perfectly acceptable, as only the message changes, not the commit contents.

Corgibytes.Freshli.Cli/CommandOptions/CheckoutHistoryCommandOptions.cs
```diff
@@ -4,6 +4,6 @@ public class CheckoutHistoryCommandOptions : CommandOptions
 {
     public string RepositoryId { get; set; }
     public string Sha { get; set; }
-    public string GitPath { get; set; }
+    public string GitPath { get; set; } = "git";
```
CodeMouse92, line 7: We do NOT want to provide this here. `System.CommandLine` should be providing the default value, which is similarly defined in `Commands/Git/GitCloneCommand.cs:20`. This is necessary for the `--help` documentation to inform the user of the default, and of course we don't want to define twice.

Corgibytes.Freshli.Cli/CommandRunners/Git/CheckoutHistoryCommandRunner.cs
```diff
@@ -1,18 +1,32 @@
...
 namespace Corgibytes.Freshli.Cli.CommandRunners.Git;
 
 public class CheckoutHistoryCommandRunner : CommandRunner<CheckoutHistoryCommandOptions>
 {
-    public CheckoutHistoryCommandRunner(IServiceProvider serviceProvider, Runner runner) : base(serviceProvider, runner)
+    private readonly GitArchive _gitArchive;
+
+    public CheckoutHistoryCommandRunner(IServiceProvider serviceProvider, Runner runner, GitArchive gitArchive) : base(serviceProvider, runner)
     {
+        _gitArchive = gitArchive;
```
CodeMouse92, line 16: I'm trying to understand this part. We are passing a `gitArchive`, and then populating it with data in `Run`. Is there a particular reason we don't just create a new `GitArchive` object here? (There may well be, I'm just looking for more insight.)

CodeMouse92 (dismissed):

Corgibytes.Freshli.Cli/CommandRunners/Git/CheckoutHistoryCommandRunner.cs
edwinkortman: Good point. I think I've made a brain fart. My intent is to separate the thing that creates the Git Archive from the command runner but we don't need an extra object for that.

edwinkortman:
This PR is going to clash again with https://github.com/corgibytes/freshli-cli/pull/112 so I suggest we wait with merging main into this until #112 is done.

Corgibytes.Freshli.Cli/Functionality/Git/GitArchiveProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+using System.IO.Compression;
+
+namespace Corgibytes.Freshli.Cli.Functionality.Git;
+
+public class GitArchiveProcess : IGitArchiveProcess
+{
+    public string Run(GitSource gitSource, GitCommitIdentifier gitCommitIdentifier, string gitPath, DirectoryInfo cacheDirectory)
+    {
+        // If it exists, make sure to empty it so we are certain we start with a clean slate.
+        var gitSourceTarget = new DirectoryInfo(cacheDirectory.FullName + "/histories" + "/" + gitSource.Hash + "/" + gitCommitIdentifier);
```
mscottford, line 12: We should be using `Path.Combine` here:

```suggestion
        var gitSourceTarget = new DirectoryInfo(Path.Combine(cacheDirectory.FullName, "histories", gitSource.Hash, gitCommitIdentifier));
```

Corgibytes.Freshli.Cli/Functionality/Git/GitArchiveProcess.cs
```diff
@@ -0,0 +1,47 @@
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true
+            }
+        };
+        archiveProcess.Start();
+        archiveProcess.WaitForExit();
+
+        if (archiveProcess.ExitCode != 0)
+        {
+            throw new GitException($"Git encountered an error:\n{archiveProcess.StandardError.ReadToEnd()}");
```
mscottford, line 38: Should we have a translation string for this?

mscottford (changes requested):
This is looking good. I just have a few requested changes. Almost all of them are related to the use of `/` as the path separator when we should be taking a platform-neutral approach to path separators instead.

features/checkout_history.feature
```diff
@@ -0,0 +1,13 @@
...
+        And a directory named "~/.freshli/histories/6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589/75c7fcc7336ee718050c4a5c8dfb5598622787b2" does not exist
+        When I run `freshli git checkout-history 6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589 75c7fcc7336ee718050c4a5c8dfb5598622787b2`
+        Then a directory named "~/.freshli/repositories/6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589" is not empty
+        And a directory named "~/.freshli/histories" exists
+        And a directory named "~/.freshli/histories/6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589/75c7fcc7336ee718050c4a5c8dfb5598622787b2" is not empty
+        And a file named "~/.freshli/histories/6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589/75c7fcc7336ee718050c4a5c8dfb5598622787b2/archive.zip" does not exist
+        And the output should contain:
+        """
+        .freshli/histories/6a2c5b97bc5113bda4845c09637043aef3d499f5b62eb252314a6bbcc7afd589/75c7fcc7336ee718050c4a5c8dfb5598622787b2
+        """
```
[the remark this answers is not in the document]
CodeMouse92: Does Ruby not work around that? I know Python does.

Corgibytes.Freshli.Cli/CommandRunners/Git/CheckoutHistoryCommandRunner.cs
edwinkortman: So I was looking at it, not quite figuring out why I did that but now I know.
It's indeed following the builder pattern. I don't want the way to create a GitArchive, or at least, the way of gathering all the necessary information tied to an object specific to the command line (`CheckoutHistoryCommandRunner`). If we decide to switch the user interface or the architecture where we don't use the CommandRunner as outward facing, then the refactor (in this instance) becomes more of a hassle. Same goes if we want to add an interface, or add architecture next to command runners, we need to think about what data to collect to create an archive. 

I'd agree that in this instance the services needed to create an archive are minimal but I find it good practice nonetheless.

edwinkortman: It is the Builder pattern, but it's value is a bit lost. 
There's two objects in play here `GitArchive` and `GitArchiveProcess`. I separated the two because I wanted to unit test the `GitArchive`. It has a test, but that's testing it returns a string. 

I'm going to merge the two. :-) Thanks for pointing it out

edwinkortman:
> I'm going to pick on commit messages for a moment. Most of these are reasonable (given that I hadn't shared about commit messages yet), but could you possibly change the one that is simply `wiring` to be more descriptive? `git rebase -i HEAD~7` should let you select that one for rewriting the message; this is one case where rewriting history should be perfectly acceptable, as only the message changes, not the commit contents.
> 
> Proper warning, I'll be a lot more nit-picky about commit messages in later reviews. I've learned from experience that these are one of the most important communication artifacts we leave. I recommend [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and [Angular Commit Message Types](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) (as these add a lot of value for minimal effort), but at minimum, we need good [commit descriptions](https://corgibytes.com/blog/2019/03/20/commit-messages/).

@CodeMouse92 Unfortunately I did the merge before the rebase so I can't rewrite it without a lot of headaches. I'll do better next time though.

features/checkout_history.feature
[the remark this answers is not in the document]
edwinkortman: I've added an issue for it. #132
I don't know but there's plenty of tests that will likely fail on Windows if Ruby does not work around it.

PR #118, opened 2022-07-05
Lint trial

edwinkortman:
Ping @CodeMouse92


CodeMouse92:
We should probably figure out how to configure this into lint.rb tomorrow before proceeding.

Issue #51, opened 2022-04-28
Add `git clone` command


mscottford:

CodeMouse92:
- [ ] Add `git` command
- [ ] Add `git clone` subcommand
- [ ] Add `repo-url` argument to `git clone` subcommand
- [ ] Add `--branch` and `--git-path` options to `git clone` subcommand
- [ ] Add function for creating or retrieving directory from cache
- [ ] Implement `GitRepository` table in cache db
- [ ] Implement `git clone` subcommand
- [ ] Write unit tests for `git`, `git clone` and the options/arguments
- [ ] Write Cucumber acceptance tests for happy paths and expected failures.

mscottford:
I forgot to mention this before. But I did some work quite a while ago over in the `freshli-lib` repo that [experimented with wrapping calls to the `git` CLI](https://github.com/corgibytes/freshli-lib/pull/243). I'm not sure if looking at that work would be useful here or not. (Of course, I realize I'm writing this when all but one of the items in the task list has been completed, so it's very likely that this comment is arriving too late to be useful.) 

The most significant part of that effort was the use of the [CliWrap library](https://github.com/Tyrrrz/CliWrap) for invoking the child command line program.

CodeMouse92:
> I forgot to mention this before. But I did some work quite a while ago over in the `freshli-lib` repo that [experimented with wrapping calls to the `git` CLI](https://github.com/corgibytes/freshli-lib/pull/243). I'm not sure if looking at that work would be useful here or not. (Of course, I realize I'm writing this when all but one of the items in the task list has been completed, so it's very likely that this comment is arriving too late to be useful.)
> 
> The most significant part of that effort was the use of the [CliWrap library](https://github.com/Tyrrrz/CliWrap) for invoking the child command line program.

Addressed in #119.

PR #119, opened 2022-07-05
Use CliWrap instead of System.Diagnostics

CodeMouse92:
Employs CliWrap, as per comment in #51.


Corgibytes.Freshli.Cli/Functionality/Git.cs
```diff
@@ -100,69 +100,66 @@ private void Delete()
...
+                    .Add("clone")
+                    .Add(Url)
+                    .Add('.')
+            )
+            .WithValidation(CommandResultValidation.None)
+            .WithWorkingDirectory(Directory.FullName)
+            .WithStandardErrorPipe(PipeTarget.ToStringBuilder(stdErrBuffer));
+
+        var task = command.ExecuteAsync().Task;
+        task.Wait();
```
mscottford, line 115: Calling `Result` on the next line implies `Wait()`, but I see little harm in being explicit here.

mscottford (approved):
This looks good!

Corgibytes.Freshli.Cli/Functionality/Git.cs
```diff
@@ -100,69 +100,66 @@ private void Delete()
...
+        var stdErrBuffer = new StringBuilder();
+        var command = CliWrap.Cli.Wrap(gitPath).WithArguments(
+                args => args
+                    .Add("clone")
+                    .Add(Url)
+                    .Add('.')
+            )
+            .WithValidation(CommandResultValidation.None)
+            .WithWorkingDirectory(Directory.FullName)
+            .WithStandardErrorPipe(PipeTarget.ToStringBuilder(stdErrBuffer));
```
mscottford, line 112: We need to add `using` here to make sure that the `CommandTask<T>` instance is properly cleaned up, because it implements `IDisposable`.

```suggestion
        using var command = CliWrap.Cli.Wrap(gitPath).WithArguments(
                args => args
                    .Add("clone")
                    .Add(Url)
                    .Add('.')
            )
            .WithValidation(CommandResultValidation.None)
            .WithWorkingDirectory(Directory.FullName)
            .WithStandardErrorPipe(PipeTarget.ToStringBuilder(stdErrBuffer));
```

Corgibytes.Freshli.Cli/Functionality/Git.cs
```diff
@@ -100,69 +100,66 @@ private void Delete()
...
-                RedirectStandardOutput = true,
-                RedirectStandardError = true
-            }
-        };
-        checkoutProcess.Start();
-        checkoutProcess.WaitForExit();
-
-        if (checkoutProcess.ExitCode != 0)
+        var stdErrBuffer = new StringBuilder();
+        var command = CliWrap.Cli.Wrap(gitPath).WithArguments(
```
mscottford, line 127: `using` is also needed here.

Corgibytes.Freshli.Cli/Functionality/Git.cs
```diff
@@ -100,69 +100,66 @@ private void Delete()
...
-                RedirectStandardOutput = true,
-                RedirectStandardError = true
-            }
-        };
-        pullProcess.Start();
-        pullProcess.WaitForExit();
-
-        if (pullProcess.ExitCode != 0)
+        var stdErrBuffer = new StringBuilder();
+        var command = CliWrap.Cli.Wrap(gitPath).WithArguments(
```
mscottford, line 148: `using` is also needed here.

mscottford (changes requested):
Eek! I noticed something after I'd already submitted an "approved" review. Sorry for the confusion.

Corgibytes.Freshli.Cli/Functionality/Git.cs
CodeMouse92: It *wasn't* implying it, which is why I mentioned in Slack about the race condition.

Corgibytes.Freshli.Cli/Functionality/Git.cs
CodeMouse92: Does not work. If you make this change in Jetbrains, it will immediately alert you with the following:

```
Type 'CliWrap.Command' used in a 'using' declaration must be implicitly convertible to 'System.IDisposable'|
```

mscottford: I liked the solution that you came up with. Thanks!

mscottford (approved):
Approved pending conflict resolution with the `main` branch.

PR #121, opened 2022-07-12
Compute history

### LINKED ISSUE #52: Add `git compute-history` command
## `git compute-history`

Used to examine the history of the specified repository and determine the sha hashes that will need to be checked out to complete a historical analysis at the intervals specified.

```
freshli [global options] git compute-history [command options] <respository-id>
```

**What this command does?**

Creates a row in the repository_histories table for each history interval stop and records the sha hash that is detected for that point in time. Each date/time and sha hash pair is also output to STDOUT on a single line per row.

Global Options

- `--cache-dir <path>`

Command Options

- `--git-path`
    - Path to the `git` binary
    - default value: `git`
   
- `--commit-history`
    - Analyzes only the points in time when files have changed in the commit history. This can be combined with `--history-invertal` to make sure that every commit is included in the result set. If `--history-interval` is not used, then only the values from each commit will be present in the output, meaning that the data will show the metrics at the time that the files were changed, but will not show how the metrics changed between commits.
- `--history-interval <duration>`
    - As the `analyze` command moves backwards in time through the history of the project, what time interval should it use to determine the points in time that are sampled
    - default value: `1m`
    - valid values: `/\d+[ymwd]?/`
        - `y` suffix – specifies duration in years
        - `m` suffix – specifies duration in months
        - `w` suffix – specifies duration in weeks
        - `d` suffix – specifies duration in days
    - When walking backward in time the following approach will be taken: given current time `T`, subtract the smallest amount of time to produce time `T'` such that `T'` when be an even interval of the specified duration suffix.
        - Example:
            - Assume
                - `T = 2022-02-01 23:11:43Z`
                - Interval duration — `1d`
                - Timezone offset — `-04:00`
                - Then `T'` would be set to `2022-02-01 00:00:00-04:00`
                - The value of `T'` would be decremented by one calendar day until the earliest commit date is reached

edwinkortman:
Fixes #52


Corgibytes.Freshli.Cli/Functionality/Git/ComputeHistory.cs
```diff
@@ -0,0 +1,63 @@
...
+            // Grouped by month
+            "month" => commitHistory.GroupBy(commit => commit.CommittedAt.Month),
+
+            // Grouped by year
+            "year" => commitHistory.GroupBy(commit => commit.CommittedAt.Year),
+            _ => null
+        };
+
+        // Per group, select the most recent commit.
+        // So if grouped per day, pick the commit committed last on that day.
```
CodeMouse92, line 53: I might be misreading something, but doesn't the LINQ statement below *always* select the commit committed last on the day, regardless of the selected history interval?

Corgibytes.Freshli.Cli/Functionality/Git/GitCommit.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Corgibytes.Freshli.Cli.Functionality.Git;
+
+public class GitCommit
+{
+    public readonly DateTimeOffset CommittedAt;
+    public readonly string ShaIdentifier;
```
CodeMouse92, line 8: Love it or hate it, the convention I've seen throughout C# is to favor get-only properties over public readonly member variables.

```suggestion
    public DateTimeOffset CommittedAt { get; }
    public string ShaIdentifier { get; } 
```

Corgibytes.Freshli.Cli/Functionality/Git/GitCommit.cs
```diff
@@ -0,0 +1,15 @@
...
+
+    public GitCommit(string shaIdentifier, DateTimeOffset committedAt)
+    {
+        ShaIdentifier = shaIdentifier;
+        CommittedAt = committedAt;
```
CodeMouse92, line 13: Nit: A common convention (at least in C and C++, but I've seen it elsewhere) is to assign to members in the same order as they're defined. If nothing else, it improves readability, and makes it easier to spot missing member assignments.

```suggestion
        CommittedAt = committedAt;
        ShaIdentifier = shaIdentifier;
```

Corgibytes.Freshli.Cli/Functionality/Git/HistoryIntervalStop.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Corgibytes.Freshli.Cli.Functionality.Git;
+
+public class HistoryIntervalStop
+{
+    public HistoryIntervalStop(DateTimeOffset committedAt, string gitCommitIdentifier)
```
CodeMouse92, line 7: I don't know if this matters, but this is semantically the opposite as the constructor for `GitCommit`: there, the parameters are "identifier, date", and here they're "date, identifier". I recommend picking one convention (identifier or date first?) and sticking to it to reduce guesswork.

features/compute_history.feature
```diff
@@ -0,0 +1,53 @@
+Feature: Compute history
```
CodeMouse92, line 1: Nice tests. :)

CodeMouse92 (changes requested):
Nice PR. I've added a few comments.

Also, thanks for the good commit messages. :)

Corgibytes.Freshli.Cli.Test/Functionality/Git/ComputeHistoryTest.cs
```diff
@@ -0,0 +1,166 @@
...
+                        "75c7fcc7336ee718050c4a5c8dfb5598622787b2"
+                    ),
+                    new(
+                        new(2021, 5, 19, 15, 24, 24, TimeSpan.Zero),
+                        "583d813db3e28b9b44a29db352e2f0e1b4c6e420"
+                    )
+                },
+                "day"
+            }
+        };
```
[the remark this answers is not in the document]
edwinkortman: I agree, it would turn it on for the entire project though (currently that means changes in 33 files).
I think we should pick this up in a separate issue. What do you think?

For the record. The settings we want are (in `.editorconfig`):
```
csharp_object_creation_when_type_evident = explicitly_typed
csharp_object_creation_when_type_not_evident = target_typed
csharp_style_implicit_object_creation_when_type_is_apparent = false
```

Corgibytes.Freshli.Cli/Functionality/Git/ComputeHistory.cs
edwinkortman: Oops, no but the variable wasn't renamed yet.

Corgibytes.Freshli.Cli.Test/Functionality/Git/ComputeHistoryTest.cs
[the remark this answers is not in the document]
CodeMouse92: Sounds reasonable.

Issue #129, opened 2022-07-13
Let `AgentsReader` return mock data from a script


edwinkortman:
Currently the `AgentsReader` is returning mock data based on whatever file path is given. This should be refactored so that it calls a shell script (or something else) that will give us some mock data to work with. That would be more realistic then it is now, and reusable for the tests as well.

PR #130, opened 2022-07-13
Create mock freshli agent in Ruby

mscottford:
Implements #129.


mscottford:
@edwinkortman Do you think this meets the needs that you identified in #129 or is there still more work to be done in your opinion?

edwinkortman:
> @edwinkortman Do you think this meets the needs that you identified in #129 or is there still more work to be done in your opinion?

No, this looks good! I can work with this. Nice work everyone :-)

Issue #124, opened 2022-07-12
`MainCommand.cs` keeps being edited by the formatter


edwinkortman:
See this comment https://github.com/corgibytes/freshli-cli/pull/112#discussion_r919069374
![image](https://user-images.githubusercontent.com/7047894/178561870-de05b166-7605-4874-803e-86edd0d038cc.png)

For some reason, the linter see it as okay and then the formatter edits it anyway or vice versa. Investigate why this is happening.

CodeMouse92:
The change is being made by `dotnet jb cleanupcode`, but flagged by `dotnet format --verify-no-changes`.

The quick fix is to run `dotnet format` *after* `dotnet jb cleanup` in `bin/format.rb`, so the order of lint checking and the order of format changing matches. This is probably a good idea anyway.

However, the above fix only works because it'll make and revert the change to `MainCommand.cs`. We will still want to figure out what the conflicting rules in `.editorconfig` are.

CodeMouse92:
It would seem that `dotnet verify` is flagging this because of this rule:

```
csharp_new_line_before_open_brace = all
```

However, I cannot seem to make Jetbrains analyzecode/cleanupcode match, even with changing the analogous rules:

```
resharper_brace_style = next_line
resharper_csharp_brace_style = next_line
resharper_csharp_initializer_braces = next_line
```

The setting continues to thrash, even after cleaning the cache.

PR #138, opened 2022-07-19
Match Formatter and Linter Precedence Order

CodeMouse92:
Initially addresses #124.


edwinkortman (approved):
Good investigation. Weird that the settings are not respected. :/

PR #140, opened 2022-07-20
Abstract External Process Calls

CodeMouse92:
Adds the Invoke.Command() and Invoke.Freshli() methods for
executing external processes with CliWrap. This eliminates
boilerplate from the rest of the project.

Changes all prior direct invocations of processes with CliWrap
to use Invoke.Command() instead.

Invoke.Freshli() method will be needed in upcoming work. This just
defines it, since it was written as part of a now-abandoned task
and worth retaining.

---

I credited Dona on this task because she helped write the method
that I further developed here into Invoke.Freshli() and Invoke.Command().


CodeMouse92:
Handing this off.

edwinkortman:
Thanks @CodeMouse92 !

PR #143, opened 2022-07-22
Disable omitting class name from constructor

edwinkortman:
This PR is without issue because it comes from this discussion:
https://github.com/corgibytes/freshli-cli/pull/121#discussion_r924735640


Corgibytes.Freshli.Cli.Test/CommandOptions/ScanCommandOptionsTest.cs
```diff
@@ -127,7 +127,7 @@ public ScanCommandOptionsTest(ITestOutputHelper output) : base(output)
         IList<OutputStrategyType> expectedOutput)
     {
         var cmBuilder = Program.CreateCommandLineBuilder();
-        Parser parser = new(cmBuilder.Command);
+        Parser parser = new Parser(cmBuilder.Command);
```
mscottford, line 130: This should become:

```suggestion
        var parser = new Parser(cmBuilder.Command);
```

mscottford (changes requested):
I'm okay with the direction of this change because the `new(new(new(` that we've seen in a few places has been confusing. But I would like to see us use `var` in cases where there would otherwise be `TypeName value = new TypeName`.

Corgibytes.Freshli.Cli.Test/CommandOptions/ScanCommandOptionsTest.cs
mscottford: I've addressed this in recent commits.

mscottford (approved):

donabp (approved):
everything looks good and clean, it's ready for a merge! :)

mscottford:
recheck

mscottford:
recheck

PR #154, opened 2022-07-25
finished removing agent reader mock data

donabp:


Corgibytes.Freshli.Cli/Services/AgentReader.cs
```diff
@@ -9,31 +11,35 @@ namespace Corgibytes.Freshli.Cli.Services;
...
-        "/usr/local/agents/bin/freshli-agent-csharp" => new(),
-        "/home/freshli-user/bin/agents/freshli-agent-ruby" => new(),
-        _ => new()
-    };
-
-    private static List<Package> ReturnListOfPackages(PackageURL packageUrl) =>
-        packageUrl.Name switch
+        var package = new List<Package>();
+        var executablePath = Directory.GetCurrentDirectory() + Path.DirectorySeparatorChar + "bin" +
+                             Path.DirectorySeparatorChar + "freshli-agent-mock.rb";
```
edwinkortman, line 18: The executable path is actually passed as parameter `string agentExecutable`

bin/freshli-agent-mock.rb
```diff
@@ -15,7 +15,7 @@
 
 case ARGV[0]
 when 'validating-package-urls'
-  puts 'pkg:nuget/org.corgibytes.flyswatter/flyswatter'
+  puts 'pkg:nuget/org.corgibytes.flyswatter/flyswatter@1.1.0 1990-01-29T12:15:25-1:00'
```
edwinkortman, line 18: I think for the tests to keep working we probably have to add the other two as well:
```
pkg:nuget/org.corgibytes.calculatron/calculatron@21.3 2022-10-16T00:00:00+0:00
pkg:nuget/org.corgibytes.calculatron/calculatron@14.6 2019-12-31T00:00:00+0:00
```

Corgibytes.Freshli.Cli/Services/AgentReader.cs
```diff
@@ -9,31 +11,35 @@ namespace Corgibytes.Freshli.Cli.Services;
...
+        while (process != null && !process.StandardOutput.EndOfStream)
+        {
+            var fileData = process.StandardOutput.ReadLine()?.Split(" ");
+            if (fileData != null)
+            {
+                package.Add(new(new(fileData[0]), DateTimeOffset.Parse(fileData[1])));
+            }
+        }
+
+        process?.WaitForExit();
```
edwinkortman, line 40: Jason added an abstraction for external CLI calls in https://github.com/corgibytes/freshli-cli/pull/140
We can use this to call the agent executable (I think you can pass "." as working directory). 
Then the output you get from that command can be processed like so:

```csharp
var packageUrlsWithDate = // Output from CLI Wrapped call
var packages = new List<Package>();

foreach (var packageUrlAndDate in packageUrlsWithDate)
{
    var separated = packageUrlAndDate.Split(" ");
    packages.Add(new(
        separated[0],
        DateTimeOffset.ParseExact(separated[1], "yyyy'-'MM'-'dd'T'HH':'mm':'ssK", null)
    ));
}
```

edwinkortman (changes requested):
Nice work! :D We can improve some small things that were merged earlier today.

Corgibytes.Freshli.Cli/Services/AgentReader.cs
donabp: why do we have to use the CliWrap if the process does the same thing? :)

edwinkortman: That's because of this PR: https://github.com/corgibytes/freshli-cli/pull/119
And this issue: https://github.com/corgibytes/freshli-cli/issues/51

For consistency I would prefer to stick to one method over introducing another way of talking to the CLI.

mscottford (approved):
Looks great. Reviewed in ensemble session with @mscottford, @donabp, and @klhoot.

Issue #135, opened 2022-07-19
[EPIC] Implement an CQRS-Lite architecture


edwinkortman:
# What's this?

One of the struggles we noticed when building the CycloneDX-based pipeline is scalability. Our original plan was to let a command spawn x amount of workers for another command. In practice, though, this is not working out as intended. We want to limit the number of workers spawned to whatever value passed as the argument, but if one command generates another command that also has a worker, they will multiply. 

After some discussion, we decided to try out an architecture loosely based on CQRS: 
https://www.eventstore.com/cqrs-pattern
https://matthiasnoback.nl/2015/01/a-wave-of-command-buses/
https://matthiasnoback.nl/2015/01/some-questions-about-the-command-bus/

# The Architecture
We can re-use much of the internal commands described in #55 but they'll need a small refactor to be able to work with a message bus. First let's dive into what it means.

With this setup instead of having Freshli CLI commands call other commands as if it were a CLI application calling another CLI application, we'll change this to, eventually, asynchronous message bus. The benefits of this setup is that we'll have an easier time scaling it horizontally because we can add more workers processing the messages. A downside is that we'll add a bunch of classes to make things work. 

**What are we going to add?**

We can make a distinction between four types of objects. Messages, Message Handlers, Events, and Event Subscribers. Another name for Message is Command but there's a clash with CLI commands, and it's runners. To avoid this we'll stick to messages. A message is a short "do this" statement. We'll get to the examples soon, so bear with me for now. This message only contains the necessary information for the Handler to do it's job. Think of just passing primitive types, and identifiers. At no point do you want to put an entire Order object into this. This is because the messages get send to the queue, and passing the queue a wall of text is going to slow things down or block the stream entirely. 

So we now have two types of objects. Message and Message Handler. These are connected, and there's only ever one handler for a particular message. An example:

```
class SetOrderToPaidMessage
{
      public OrderId orderId;
      public DateTime paidAt;
}
```

```
class SetOrderToPaidMessageHandler
{
    public __invoke(SetOrderToPaidMessage $message) 
    {
         // Do things
    }
}
```

**How are they connected?**

We'll be running workers. These workers sole responsibility is to offer messages from the message bus to the handlers. The handlers will then let the workers know if they can or cannot process the message. 

**What about Events?**

[...]

Issue #55, opened 2022-04-29
[EPIC] Implement CycloneDX-based pipeline


mscottford:
# What’s this?

A re-imagining of Freshli that takes advantage of the ecosystem of tools that generate CycloneDX BOM files. The current idea is to drive this from the Freshli CLI.

# Architecture

The CLI is going to be split into user-facing commands and internal commands. This is similar to the approach that’s taken by the `git` set of tools. In that ecosystem, there are a set of commands that are meant to be user-facing (called porcelain commands), and there are other commands that are in support of internal operations and other commands (called plumbing commands).

We’ll do our best to follow the [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) for building out these commands:

- Do one thing, and do it well
- The output of each command can be used as input for another command, with the ability to read information from STDIN and write information to STDOUT - sometimes this point will be deviated from when it makes more sense to persist output to the file system so that it can be used by multiple programs reading the same information at the same time

One reason for taking this approach is that it allows us to use the operating system’s process scheduling to get parallelization of activities. Another reason is that it allows us to interact with tools that we did not write, such as `git` for source code integration or `mvn` for build system integration or a command line tool that knows how to generate or manipulate CycloneDX BOM files. 

It’s expected that many of the commands below will permit running multiple workers, with work to be completed stored in queue tables in the local database. The local database will also be used to cache information from various external sources and to cache the results of expensive computations.

One goal with caching is to make the analyze command run quickly against commits that it has already analyzed. This would be particularly useful when running the CLI from a CI environment, since most of them support storing cached information between builds.

```mermaid
flowchart LR
    start(analyze start)
    subgraph cache
        prepare
    end
    subgraph git
        clone --> compute-history
        compute-history --> checkout-histories
        checkout-histories --> compute-remaining-histories{{Histories\nRemain?}}
        compute-remaining-histories -->|Yes| checkout-history
        checkout-history --> compute-remaining-histories
        compute-remaining-histories -->|No| history-done(Done)
    end
    
    subgraph bom
        generate(generate start) --> generate-histories

        generate-histories --> generate-remaining-histories{{Histories\nRemain?}}
        generate-remaining-histories -->|Yes| generate-history
        generate-history --> generate-remaining-histories
        generate-remaining-histories -->|No| generate-done(Done)				
    end
    subgraph compute
        libyear
    end
    subgraph persist
        save-to-database
    end
    subgraph output
        create-local-report
    end
    subgraph publish
[...]

PR #156, opened 2022-07-26
Adds an "application engine" for processing application activities

### LINKED ISSUE #137: Add a Message Bus
Following the architecture in this ticket: https://github.com/corgibytes/freshli-cli/issues/135
We need something to deliver our Messages and Events. Investigate which options we have in C# for a synchronous message bus. 

**Open questions:**
- [ ] Can we make busses specific for messages, and events?
- [ ] How's error handling done?
- [ ] Can we retry failed messages?

Some resources:
https://docs.microsoft.com/en-us/azure/architecture/guide/architecture-styles/event-driven
https://docs.microsoft.com/en-us/azure/architecture/patterns/publisher-subscriber
https://docs.microsoft.com/en-us/azure/architecture/patterns/category/messaging
https://docs.microsoft.com/en-us/dotnet/api/system.messaging.messagequeue?view=netframework-4.8

edwinkortman:
Fixes #137

This pull request introduces an "application engine", implemented using the Hangfire library. This engine manages a queue that contains "application activities" and "applications events". An activity is capable of firing an event, and an event is capable of dispatching an activity.

The queue is serviced by a configurable number of background worker threads, permitting parallel execution and processing of activities and events.

This architecture is going to be used in place of the one that we've been working towards so far, where command line commands execute other command line commands and process their results. Instead, `IApplicationActivity` and `IApplicationEvent` objects are going to be created to process the pipeline of work that has been sketched out in #55.

This commit is the first step towards implementing something similar to the approach that is is outlined and proposed in #135. The exact implementation here differs slightly from what was proposed, but the spirit of the implementation is much the same.

This commit contains code that was copy-and-pasted from the Hangfire.AspNetCore library. The classes that are imported this way are responsible for setting up dependency injection and logging support.

Since Hangfire depends on Newtonsoft.Json, a custom class has been added to assist with the integration between the Json deserialization and the dependency injection framework that we're using. This allows the deserialization functionality to refer to object instances that are present in the inversion of control container that's managed by the dependency injection framework.

This giant pull request represents a group effort.


edwinkortman:
From the ensemble session:
- Look into Hangfire https://www.hangfire.io/
- Currently there's no need for persistence

Corgibytes.Freshli.Cli/Functionality/Message/MessageBus.cs
```diff
@@ -1,27 +1,31 @@
...
-        _connectableObservable.Connect();
+        _jobClient = jobClient;
         SubscribeHandlers();
     }
 
-    // ReSharper disable once UnusedMember.Global
-    public void Dispatch(IMessage job) => _jobs.OnNext(job);
+    public void Dispatch(IMessage job)
+    {
+        BackgroundJob.Enqueue(() => new TestMessageHandler().Handle(job));
```
edwinkortman, line 19: This is just for testing purposes to get Hangfire working, because the dispatch shouldn't be aware of the handler. It should throw the message into the queue and be done with it.

Corgibytes.Freshli.Cli/IoC/FreshliServiceBuilder.cs
```diff
@@ -114,5 +116,9 @@ private void RegisterComputeLibYearCommand()
         Services.AddTransient<IAgentReader, AgentReader>();
     }
 
-    private void RegisterMessageBus() => Services.AddSingleton<MessageBus>();
+    private void RegisterMessageBus()
+    {
+        Services.AddSingleton<MessageBus>();
+        Services.AddHangfire();
```
edwinkortman, line 122: This doesn't work because there's no definition for `AddHangfire`. I've checked out the Hangfire repository, and it's a extension method that should be loaded somehow.

mscottford:
Code review was performed by an ensemble team of @mscottford, @klhoot, and @donabp.

PR #158, opened 2022-07-27
CLA bot

edwinkortman:
To follow the examples set by Freshli Lib and Freshli Java agent, let this project store it's CLA signatures in the same repository and retrieve the document from the same repository. 

I'm not merging the two signature files because I think we have to sign it again, which makes sense because the document is in a different place so it could've been changed (it hasn't).


.github/workflows/cla.yml
```diff
@@ -18,14 +18,11 @@ jobs:
           PERSONAL_ACCESS_TOKEN : ${{ secrets.CLA_ACCESS_TOKEN }}
         with:
           path-to-signatures: 'CLA/signatures/v1.0/cla.json'
-          path-to-document: 'https://github.com/corgibytes/freshli-cli/blob/main/CLA/CLA-v1.0.md'
+          path-to-document: 'https://github.com/corgibytes/freshli/blob/main/CLA/CLA-v1.0.md'
```
mscottford, line 21: No such file exists at this location. Do we need to add it there?

.github/workflows/cla.yml
```diff
@@ -18,14 +18,11 @@ jobs:
...
           branch: 'main'
           allowlist: dependabot[bot]
-
-          #Configurable inputs
-          #remote-organization-name: enter the remote organization name where the signatures should be stored (Default is storing the signatures in the same repository)
-          #remote-repository-name:  enter the  remote repository name where the signatures should be stored (Default is storing the signatures in the same repository)
+          remote-organization-name: 'corgibytes'
+          remote-repository-name: 'freshli-lib-cla'
```
mscottford, line 25: Instead of `freshli-lib-cla`, I wonder if it makes more sense to use `freshli-cla` as the repository where the CLA document and the signatures are housed. We could simply rename the `freshli-lib-cla` repository in that case.

.github/workflows/cla.yml
mscottford: I think it would be best for the CLA document to be located in the same repository as the signatures.

CLA/CLA-v1.0.md
```diff
@@ -1,24 +0,0 @@
...
-
-1.	You are not expected to provide support for Your Contributions, except to the extent You desire to provide support. You may provide support for free, for a fee, or not at all. Unless required by applicable law or agreed to in writing, You provide Your Contributions on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied, including, without limitation, any warranties or conditions of TITLE, NON- INFRINGEMENT, MERCHANTABILITY, or FITNESS FOR A PARTICULAR PURPOSE.
-
-1.	Should You wish to submit work that is not Your original creation, You may submit it to Freshli separately from any Contribution, identifying the complete details of its source and of any license or other restriction (including, but not limited to, related patents, trademarks, and license agreements) of which you are personally aware, and conspicuously marking the work as "Submitted on behalf of a third-party: [named here]".
-
-1.	You agree to notify Freshli of any facts or circumstances of which you become aware that would make these representations inaccurate in any respect.
-
-_Revised: November 23, 2020._
```
mscottford, line 24: If we're going to remove the CLA from this repository, then I think we need to add a note in the README to the location where the document can be found along with a mention that it has to be signed when submitting a pull request.

CLA/signatures/v1.0/cla.json
```diff
@@ -1,92 +0,0 @@
...
-    {
-      "name": "edwinkortman",
-      "id": 7047894,
-      "comment_id": 1154846393,
-      "created_at": "2022-06-14T07:58:05Z",
-      "repoId": 299162738,
-      "pullRequestNo": 99
-    }
-  ]
-}
```
mscottford, line 92: If we're moving in the direction of having a single repository with CLA signatures for all of the projects, then I think it makes sense to merge the signature files together. That way we have a clear match between commits and signatures. I wouldn't want to create a potential issue where there are commits that don't appear to be covered by signatures.

mscottford (dismissed):
I like this. I just have a few questions that need to be answered before we merge this in.

.github/workflows/cla.yml
edwinkortman: I agree. I copied this configuration from the Freshli Lib and didn't check the link. I'll fix it there as well. :-)

.github/workflows/cla.yml
edwinkortman: That would look cleaner :+1: I'm renaming it

CLA/CLA-v1.0.md
edwinkortman: Good point! I've added it to `CONTRIBUTING.md`. I think that makes more sense because we only want it signed if someone contributes.

CLA/signatures/v1.0/cla.json
edwinkortman: Ahh, that's a good point. I'll merge them.

Issue #147, opened 2022-07-25
Start analysis


edwinkortman:
Belongs to #135 

This message, and it's handler will save an analysis object to the database. This object contains the cache directory, repository url, repository branch, history interval, and generates an analysis identifier. This is then emitted in the Event. 

## Classes to create
- [ ] Analysis
    - [ ] Property: Analysis identifier
    - [ ] Property: Cache directory
    - [ ] Property: Repository url
    - [ ] Property: Repository branch
    - [ ] Property: History interval 
- [ ] StartAnalysisMessage
    - [ ] Param: Cache directory
    - [ ] Param: Repository url
    - [ ] Param: Repository branch
    - [ ] Param: History interval
- [ ] StartAnalysisMessageHandler
    - [ ] Generates an analysis identifier
    - [ ] Saves the analysis object to the database
- [ ] AnalysisStartedEvent
    - [ ] Analysis identifier

PR #164, opened 2022-08-03
Implements "start analysis" using application engine

mscottford:
Implements #147.

Note: This pull request applies some refactoring to the way the cache is handled so that the objects are easier to write tests for. This was needed to create adequate tests for the classes that were created.

This pull request adds a `StartAnalysisActivity` class and `RestartAnalysisActivity` class. The primary difference between the two is that the `StartAnalysisActivity` class will fire a `CacheWasNotPreparedEvent` when the cache directory has not been prepared. This creates an opportunity for the cache to then be prepared before the `RestartAnalysisActivity` is dispatched. The `RestartAnalysisActivity` fires a `UnableToRestartAnalysis` if the cache directory has not been prepared. This event is considered a "failure", and the idea is that processing events will stop when a failure event is handled.

There are some important `TODO` comments that are included in the implementation. These will need to be implemented after the `cache prepare` functionality is implemented using `Activity` and `Event` classes. #174 has been created to track that this work needs to be completed.


Corgibytes.Freshli.Cli/Functionality/Analysis/CacheWasNotPreparedEvent.cs
```diff
@@ -0,0 +1,28 @@
...
+    public IServiceProvider ServiceProvider { get; init; } = null!;
+
+    public string CacheDirectory { get; init; } = null!;
+    public string RepositoryUrl { get; init; } = null!;
+    public string? RepositoryBranch { get; init; }
+    public string HistoryInterval { get; init; } = null!;
+
+    public override void Handle(IApplicationActivityEngine eventClient) =>
+        // TODO: Actually prepare the cache before restarting the analysis
+        eventClient.Dispatch(new RestartAnalysisActivity(
```
klhoot, line 18: I was wondering if an event should have a 'side effect' like dispatching a new activity, or should this be done in an event listener?

mscottford: We chatted about this during a recent ensemble session, but I wanted to record a note of that conversation here. I think that this side effect is okay because we limit what the event can do. An event's `Handle` method can't directly fire other events, it can only dispatch activities. And similarly, for activities, an activity can't directly dispatch an activity from its `Handle` method, it's only able to fire events.

PR #169, opened 2022-08-05
added CycloneDX.Core package to verify a file is a CycloneDX file. 

### LINKED ISSUE #62: Add `agents verify` command
## `agents verify`

Used to test all of the language agents that are available for use. It tests if all the agents are adhering to the contract. e.g. does it have the expected output for `validating-repositories`. 

```
freshli [global options] agents verify [command options] [language-name]
```

**What this command does?**

When no language name is specified, delegates to the `agents detect` to detect each available for agent, and then operates as if the `agents verify` command had been called once for each available language.

When a language is specified invokes the language agent’s `validating-repositories` command to get the urls of a publicly available repositories that the agent can be validated against. Each provided repository is then cloned as a child of `CACHE_ROOT/<lang>/` directory. It is up to the language agent to determine the correct number of repositories to validate against, but the expectation is that at least one repository url is provided for each manifest file format that the language agent knows how to detect and process. If any of the repositories fails to clone, then an error is generated.

For each repository that has been checked out, the language agent’s `detect-manifests` command is invoked. The command output is then validated to ensure that at least one manifest file has been detected. If no manifest files are detected, then then an error is generated.

For each repository, the language agent’s `process-manifests` command is then invoked. It is expected that the output of the command contains the following for each manifest file that was processed:

- manifest file location in the cache tree
- resulting bom file location in the cache tree
- optionally, if a modified manifest file was generated as part of the processing, the location in the cache tree of the original manifest file

If the language agent’s `process-manifests` command does not process the same number of manifests as were detected by the language agent’s `detect-manifests` command, then an error is generated.

If the output of the `process-manifests` command lists files that don’t exist or are empty, then an error is generated.

If the resulting CycloneDX bom file is not a valid JSON-formatted CycloneDX bom file, then an error is generated.

If the any residual modifications to the cloned source code repository are detected after `process-manifests` command has completed, then an error is thrown. This is the same validation that is performed when running the `bom process-all-manifests` command with the `--verify-directories` option.

For each language that this command evaluates, the result of the verification is listed along with the number of repositories that were tested and the amount of time that it took to perform the verification.

Global Options

- `--cache-dir <path>`

Command Options

- `--workers`
    - the number of worker processes that should be running at any given time. This defaults to twice the number of CPU cores.
    - default: N+1 where N is the number of cores that are detected in the CPU

donabp:
Fixes #62.

also added logic to verify a file is a CycloneDX file


mscottford:
Let's go ahead and merge this in. We can make more changes to `agents verify` once we have actual agent implementation to evaluate it against.

PR #170, opened 2022-08-05
Explicitly set the culture to the invariant culture

### LINKED ISSUE #123: `CurrentCulture` is explicitly set to `InvariantCulture` but where?
According to the README.md, the `CurrentCulture` is explicitly set to `InvariantCulture`, but where?

https://github.com/corgibytes/freshli-cli#culture-and-language-support
> Data (such as dates and numeric formatting) are NOT localized. Dates and numeric formats use the CurrentCulture which is explicitly set to the [invariant culture](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.invariantculture?view=net-6.0).

This issue originates from here https://github.com/corgibytes/freshli-cli/pull/112#discussion_r919044544

edwinkortman:
Fixes #123 

https://stackoverflow.com/questions/12729922/how-to-set-cultureinfo-invariantculture-default
https://docs.microsoft.com/EN-US/dotnet/core/runtime-config/globalization

There are a few ways that invariant culture can be set as default yet, from what I could find, none of those were being used. 

### Before
Screenshot from 2022-08-05 15-10-37

### Beafter
Screenshot from 2022-08-05 15-11-00

### DIFF
--- a/Corgibytes.Freshli.Cli/Corgibytes.Freshli.Cli.csproj
+++ b/Corgibytes.Freshli.Cli/Corgibytes.Freshli.Cli.csproj
@@ -19,6 +19,7 @@
     <Company>Corgibytes</Company>
     <AssemblyName>freshli</AssemblyName>
     <Nullable>enable</Nullable>
+    <InvariantGlobalization>true</InvariantGlobalization>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


mscottford (approved):
Nice find!

Issue #148, opened 2022-07-25
Clone git repository


edwinkortman:
Belongs to #135 

This message, and it's handler will clone the git repository into the cache directory, and generate a repository identifier. This messages happens as a reaction to the AnalysisCreatedEvent from #147 

## Things to do
- [ ] AnalysisCreatedEvent.Handle method
    - [ ] Passes the message `CloneGitRepositoryMessage` to the message bus
- [ ] CloneGitRepositoryActivity
    - [ ] Param: Analysis identifier
- [ ] CloneGitRepositoryActivity.Handle method (This does what our current `GitCloneCommandRunner::Run` does)
    - [ ] Generates an repository identifier
    - [ ] Clone the git repository into the cache directory
- [ ] GitRepositoryClonedEvent
    - [ ] Analysis identifier
    - [ ] Repository identifier

edwinkortman:
I've only done some refactoring for this issue to prepare for the new architecture. :-)
https://github.com/corgibytes/freshli-cli/pull/171

PR #171, opened 2022-08-05
Refactor the Git Source object

edwinkortman:
I ran into a few issues that made testing with this object difficult. So as a preparation for #148 where I want to add tests, I've extracted out the logic needed to clone a git repository and put it in the repository. The object we are left with is a Cached Git Source. Now, if we want to, we can mock the repository and do whatever we want.


mscottford (approved):
This looks good to me. I'm going to merge in the `main` branch and make sure that everything still passes or fix any failures and then merge it in.

Issue #163, opened 2022-08-03
Cache Destroy


mscottford:
Refactor implementation to use the new CQRS-lite approach.

mscottford:
This was implemented by #181.

PR #181, opened 2022-08-10
Migrate cache destroy to new architecture

klhoot:
Implements #163 

Also fixed an issue with the way jobs were being dispatched that caused arguments in the activity to not be serialized as expected.


mscottford:
Assigning to @donabp for review. I know that we worked on much of it during an ensemble session, and I figure it's probably still a good idea to have someone give the whole thing a second check.

mscottford:
> Assigning to @donabp for review. I know that we worked on much of it during an ensemble session, and I figure it's probably still a good idea to have someone give the whole thing a second check.

Scratch that. I'm going to go ahead and merge this one in.

PR #204, opened 2022-08-29
Compute history

### LINKED ISSUE #149: Compute history
Belongs to #135 

This message, and it's handler are going to determine SHA hashes that need to be checked out. For each SHA hash that needs to be checked out create an event. This happens as a reaction to the git repository being cloned in #148.

## Things to do
- [ ] GitRepositoryClonedEvent.Handle
    - [ ] Passes the `ComputeHistoryActivity` to the message bus
- [x] ComputeHistoryActivity
    - [x] Param: Analysis id
    - [x] Param: Repository id
- [x] ComputeHistoryActivity.Handle method
    - [x] Call the `ComputeHistory` service
    - [x] Per `HistoryIntervalStop` found, emit Event `HistoryIntervalStopFound`
- [x] HistoryIntervalStopFoundEvent
    - [x] Param: repository identifier
    - [x] Param: commit identifier

edwinkortman:
Fixes #149


Corgibytes.Freshli.Cli/CommandRunners/Cache/PrepareCacheActivity.cs
```diff
@@ -10,14 +10,17 @@ namespace Corgibytes.Freshli.Cli.CommandRunners.Cache;
...
+        string historyInterval = "", string gitPath = "")
     {
         CacheDirectory = cacheDirectory;
         RepositoryUrl = repositoryUrl;
         RepositoryBranch = repositoryBranch;
         HistoryInterval = historyInterval;
+        GitPath = gitPath;
     }
 
+    public string GitPath { get; init; }
```
mscottford, line 22: Storing the path to get in the cache database could potentially cause problems. It would be possible (however unlikely) for someone to change the path of the git executable between runs of the `freshli` application. This would mean that the git executable no longer exists in the same location as it did before.

If this is a piece of data that needs to be passed around, perhaps it needs to exist in something like an `ISettings` or `IApplicationSettings` interface, and that could be passed around instead.

mscottford (approved):

edwinkortman:
@mscottford I've re-requested your review as I've added the implementation for AnalysisLocation and I want to know if this is how you envisioned it.

mscottford:
Most recent commit approved during pairing session.

PR #206, opened 2022-08-30
Migrate `git checkout-history` to use CQRS-lite architecture

### LINKED ISSUE #150: Checkout history
Belongs to #135 

This message, and it's handler will checkout a specific historical point for a given repository. This messages happens as a reaction to the `HistoryIntervalStopFoundEvent` from #150

## Things to do
- [ ] HistoryIntervalStopFoundEvent.Handle
    - [ ] Passes the message `CheckoutHistory` to the message bus
- [x] CheckoutHistoryActivity
    - [x] Param: AnalysisLocation
- [x] CheckoutHistoryActivity.Handle (This does what our current `CheckoutHistoryCommandRunner::Run` does)
    - [x] Checkout specific historical point
- [x] HistoryStopCheckedOutEvent
    - [x] Param: AnalysisLocation

mscottford:
Fixes #150.


Corgibytes.Freshli.Cli/Functionality/Git/IGitManager.cs
```diff
@@ -0,0 +1,7 @@
+namespace Corgibytes.Freshli.Cli.Functionality.Git;
+
+public interface IGitManager
+{
+    public GitCommitIdentifier ParseCommitSha(string commitSha);
```
edwinkortman, line 5: I think we should stick to calling it either "commit id" or "commit sha". I know they are the same thing but having a different name might confuse someone who doesn't know it's the same.

Corgibytes.Freshli.Cli/Functionality/Git/CachedGitSourceRepository.cs
```diff
@@ -6,6 +6,7 @@
 using Corgibytes.Freshli.Cli.DataModel;
 using Corgibytes.Freshli.Cli.Exceptions;
 using Corgibytes.Freshli.Cli.Resources;
+using Newtonsoft.Json;
```
edwinkortman, line 9: In an earlier PR we wanted to move away from Newtonsoft.Json. Do we still want to?
I did a bit of googling for how to achieve the same thing in System.Text.Json and according to this stackoverflow you need JsonInclude for private properties:
```csharp
[System.Text.Json.Serialization.JsonInclude]
[System.Text.Json.Serialization.JsonPropertyName("animals")]
public string AnimalForSerializer
{
    get => AnimalForMe.ToString() + "s";
    private set => AnimalForMe = Enum.Parse<Animal>(value.TrimEnd('s'));
}
```

https://stackoverflow.com/questions/58475629/what-is-the-equivalent-of-newtonsoft-jsons-json-nets-jsonproperty-field-in-s

.devcontainer/Dockerfile
```diff
@@ -1,8 +1,39 @@
 # See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.234.0/containers/dotnet/.devcontainer/base.Dockerfile
 
+### Install .NET 6.0 SDK
+# copied from https://github.com/microsoft/vscode-dev-containers/blob/9ead9dfe607ac62c408180d25a71d499c0876e39/containers/dotnet/.devcontainer/base.Dockerfile
 # [Choice] .NET version: 6.0, 5.0, 3.1, 6.0-bullseye, 5.0-bullseye, 3.1-bullseye, 6.0-focal, 5.0-focal, 3.1-focal
-ARG VARIANT="6.0-bullseye"
-FROM mcr.microsoft.com/vscode/devcontainers/dotnet:0-${VARIANT}
```
edwinkortman, line 5: I like that we no longer depend on an intermediate image and that each step has it's own chapter for what it needs when installing. :D

bin/format.rb
```diff
@@ -10,7 +10,8 @@
 status = execute('bundle check > /dev/null')
 status = execute('bundle install') unless status.success?
 
-resharper_status = execute('dotnet jb cleanupcode freshli-cli.sln') if status.success?
+resharper_status = execute('dotnet tool restore') if status.success?
+resharper_status = execute('dotnet jb cleanupcode freshli-cli.sln') if resharper_status.success?
```
edwinkortman, line 14: hah, whoops. Nice catch!

edwinkortman (changes requested):
Just a few comments! Looks good :+1:

Corgibytes.Freshli.Cli/Functionality/Git/IGitManager.cs
mscottford: Good point. I'll clean that up.

mscottford: This was taken care of in 12bca43 and 031ee33.

Corgibytes.Freshli.Cli/Functionality/Git/CachedGitSourceRepository.cs
mscottford: I would love to get rid of Newtonsoft.Json, but it's used by Hangfire. I don't think the System.Text.Json attributes will work the same way.

Corgibytes.Freshli.Cli.Test/Functionality/Analysis/AnalysisLocationTest.cs
```diff
@@ -0,0 +1,32 @@
...
+    [Fact]
+    public void PathWithoutCommitId()
+    {
+        var cacheDirectory = "/path/to/cache/dir";
+        var repositoryId = Guid.NewGuid().ToString();
+        var location = new AnalysisLocation(cacheDirectory, repositoryId);
+
+        var expectedPath = Path.Combine(cacheDirectory, "repositories", repositoryId);
+        Assert.Equal(expectedPath, location.Path);
+    }
```
edwinkortman, line 19: This is nice! I ran into this as well but didn't write a test for it, I also wasn't sure if `Path` should return a "totally" different path when commit id is given but I think in our app it works so I'm glad we had the same conclusion.

edwinkortman (approved):

edwinkortman:
@mscottford I'm merging this as I can use the analysislocation changes in my the analyze command branch.

Issue #208, opened 2022-08-30
Let `HistoryIntervalStopFoundEvent.Handle` pass `CheckoutHistoryActivity` to the activity engine


edwinkortman:
This spawns from https://github.com/corgibytes/freshli-cli/pull/207
Blocked by this PR: https://github.com/corgibytes/freshli-cli/pull/204/files#diff-8b3691b75312db3758ae2ac096f46f86676c6534002d0cb8832fd462933b4de0

- [x] HistoryIntervalStopFoundEvent.Handle
  - [x] Passes the message CheckoutHistoryActivity to the message bus

Issue #205, opened 2022-08-30
Let `GitRepositoryClonedEvent.Handle` pass `ComputeHistoryActivity` to the activity engine


edwinkortman:
Blocked by #148 
This comes from #149 but is dependant on #148 finished & merged first because that's the one adding the Event.

- [ ] GitRepositoryClonedEvent.Handle
    - [ ] Passes the ComputeHistoryActivity to the message bus

PR #212, opened 2022-08-31
Analyze command

### LINKED ISSUE #56: Add `analyze` command
NOTE: This issue is still in a draft state. Its contents will likely change as new commands are identified as part of the pipeline.

## `analyze`

The primary user-facing command. This command will delegate to other `freshli` commands to accomplish its work. It will manage work queues to enable parallelization.

```bash
freshli [global options] analyze [command options] <repo url>|<local dir>
```

**What this command does?**

With no options specified, performs analysis locally, and then sends the results to the Freshli web app so that the results can be viewed at at URL that will be provided in the command output.

Global Options

- `--cache-dir <path>`

Command Options

- `--git-path`
    - Path to the `git` binary
    - default value: `git`
- `--branch <name>`
    - The branch to checkout after cloning the repository
    - If the option is not specified, then no checkout command will be issued. The remote server’s default branch will be used instead.
- `--commit-history` (this might be moved to the "build later" category per #188)
    - Analyzes only the points in time when files have changed in the commit history. This can be combined with `--history-invertal` to make sure that every commit is included in the result set. If `--history-interval` is not used, then only the values from each commit will be present in the output, meaning that the data will show the metrics at the time that the files were changed, but will not show how the metrics changed between commits.
- `--history-interval <duration>` (this might be moved to the "build later" category per #187)
    - As the `analyze` command moves backwards in time through the history of the project, what time interval should it use to determine the points in time that are sampled
    - default value: `1m`
    - valid values: `/\d+[ymwd]?/`
        - `y` suffix – specifies duration in years
        - `m` suffix – specifies duration in months
        - `w` suffix – specifies duration in weeks
        - `d` suffix – specifies duration in days
    - When walking backward in time the following approach will be taken: given current time `T`, subtract the smallest amount of time to produce time `T'` such that `T'` when be an even interval of the specified duration suffix.
        - Example:
            - Assume
                - `T = 2022-02-01 23:11:43Z`
                - Interval duration — `1d`
                - Timezone offset — `-04:00`
                - Then `T'` would be set to `2022-02-01 00:00:00-04:00`
                - The value of `T'` would be decremented by one calendar day until the earliest commit date is reached
- `--workers` — the number of worker processes that should be running at any given time. This defaults to twice the number of CPU cores.

Running this command will dispatch a `StartAnalysisActivity`. Additionally, this command will be responsible for adjusting the number of application engine workers to match the provided value of the `--workers` parameter.

Tasks:

- [ ] Add Cucumber/Aruba-based acceptance tests for this command
- [x] Implement handling `analyze` command
    - [x] Adjust number of workers (this could be considered a nice to have if it turns out to be difficult)
    - [x] Pass command line parameters to `StartAnalysisActivity` (all except `--workers` should be passed in)
- [x] Remove the following commands
   - [x] `git checkout-history`
   - [x] `git compute-history`
   - [x] `git clone`
   - [x] `compute-libyear`
- [ ] Write results to freshli-web database
    - [ ]  TBD columns

edwinkortman:
Fixes #56.

So, ehm. It's a bit of a mess because I merged in https://github.com/corgibytes/freshli-cli/pull/224 and then had to make some changes to it because I forgot something there. My apologies :disguised_face:  

>  Adjust number of workers (this could be considered a nice to have if it turns out to be difficult)

So I've tried to do this but I don't think this is possible without recreating the entire Hangfire server. 
What I had hoped would work is adjust the background server job options but I couldn't confirm an increase in workers.

```csharp
public AnalyzeRunner(
        IServiceProvider serviceProvider,
        Runner runner,
        IApplicationActivityEngine activityEngine,
        ICacheManager cacheManager,
        BackgroundJobServerOptions backgroundJobServerOptions,
        IHostedService backgroundJobServer
        ) : base(serviceProvider, runner)
    {
        _activityEngine = activityEngine;
        _cacheManager = cacheManager;
        _backgroundJobServerOptions = backgroundJobServerOptions;
        _backgroundJobServer = backgroundJobServer;
    }
...

_backgroundJobServerOptions.WorkerCount = 25;
_backgroundJobServer.StopAsync(new CancellationToken());
_backgroundJobServer.StartAsync(new CancellationToken());
```


mscottford:
Let's move controlling the worker count via a parameter to a separate issue, since there were challenges getting it to work.

mscottford:
Are #205 and #208 also fixed by this pull request? If so, we should correctly note that in the description.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
```diff
@@ -0,0 +1,104 @@
...
+        string expectedHistoryInterval, int expectedWorkers
+    )
+    {
+        var cmBuilder = Program.CreateCommandLineBuilder();
+        var parser = new Parser(cmBuilder.Command);
+
+        var result = parser.Parse(args);
+
+        var gitPath = result.GetOptionValueByName<string>("git-path");
+        gitPath?.Should().NotBeEmpty().And.Be(expectedGitPath);
```
mscottford, line 78: If `gitPath` is null then there's no assertion made against it's value. That doesn't seem correct to be. Seems like it doesn't properly test it's default value.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        var branch = result.GetOptionValueByName<string>("branch");
+        branch?.Should().NotBeEmpty().And.Be(expectedBranch);
```
mscottford, line 81: Similar to the comment about `gitPath`. If the value is `null` then there's no assertion made against it's value.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        var commitHistory = result.GetOptionValueByName<bool>("commit-history");
+        commitHistory.Should().Be(expectedCommitHistory);
+
+        var historyInterval = result.GetOptionValueByName<string>("history-interval");
+        historyInterval?.Should().NotBeEmpty().And.Be(expectedHistoryInterval);
```
mscottford, line 87: Same as `gitPath` and `branch`. A `null` value for `historyInterval` means that it doesn't get asserted.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        var workers = result.GetOptionValueByName<int>("workers");
+        workers.Should().Be(expectedWorkers);
+    }
+
+    [Fact]
+    public void Send_Args_ReturnsAnalyzeArguments()
+    {
+        var cmBuilder = Program.CreateCommandLineBuilder();
```
mscottford, line 96: I'm not a huge fan of the `cm` prefix on `cmBuilder`. I think just `builder` would be more than fine in this context, or the call could be inlined and removed altogether.

Corgibytes.Freshli.Cli.Test/Functionality/CacheManagerTest.cs
```diff
@@ -32,7 +32,7 @@ public void SavePersistsACachedAnalysisAndGeneratesAnId()
 
         var cache = cacheManager.GetCacheDb(_tempCacheDir);
 
-        var expectedAnalysis = new CachedAnalysis("https://git.example.com", "main", "1m");
+        var expectedAnalysis = new CachedAnalysis("https://git.example.com", "main", "1m", true);
```
mscottford, line 35: I'm not a big fan of boolean values being passed as parameters, because it's always hard to tell what they mean. Can we either change all of these calls to look like:

```suggestion
        var expectedAnalysis = new CachedAnalysis("https://git.example.com", "main", "1m", useCommitHistory: true);
```

Or switch to using an enumeration. Something like `CommitHistory.Full` vs `CommitHistory.Minimal`, where `CommitHistory.Full` is analogous to a value of `true` and `CommitHistory.Minimal` is analogous to a value of `false`.

mscottford:
From reading through the changes, it looks like this pull request also addresses #213.

Corgibytes.Freshli.Cli/CommandRunners/AnalyzeRunner.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        _eventEngine.On<LibYearComputedEvent>(computedEvent =>
+        {
+            var libYearSummed = 0.0;
+            if (computedEvent.LibYearPackages != null)
+            {
+                libYearSummed = computedEvent.LibYearPackages.Sum(libYear => libYear.LibYear);
+            }
+
+            context.Console.Out.WriteLine($"Libyear at {computedEvent.AnalysisLocation?.CommitId} is {libYearSummed}");
```
mscottford, line 58: I like that this information is being written out. It'll help show progress as the command is running along. I think the output would be more meaningful if it included the path to the manifest that's being analyzed. Many git repositories are going to have more than one dependency manifest. In those cases, you'd get this output twice for a given CommitId. Also the date maters, too... so that should be included as well. So I think the output would need to be something like

"LibYear for {manifest} at {CommitId} on {date} is {libYearSummed}". What do you think? We can also create an issue for this and have it wait until later. I wouldn't think that this is needed before we demo.

We could also write this line to the log as `Information` if it's going to make the console output too noisy. If we write a detailed message to the log, then I think we could just write out a `.` to the console to show that progress is being made.

Corgibytes.Freshli.Cli/CommandRunners/ComputeLibYearCommandRunner.cs
```diff
@@ -40,8 +41,11 @@ public override int Run(ComputeLibYearCommandOptions options, InvocationContext
                 CliOutput.ComputeLibYearCommandRunner_Run_FilePath_should_not_be_null_or_empty);
         }
 
-        _activityEngine.Dispatch(new ComputeLibYearActivity(_calculateLibYearFromCycloneDxFile,
-            options.FilePath.ToString()));
+        _activityEngine.Dispatch(new ComputeLibYearActivity(
+            _calculateLibYearFromCycloneDxFile,
+            options.FilePath.ToString(),
+            new AnalysisLocation("this-runner-will-be-removed", "this-runner-will-be-removed")
```
mscottford, line 47: Should the runner be removed now?

Corgibytes.Freshli.Cli/CommandRunners/LoadServiceCommandRunner.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    private IApplicationActivityEngine ActivityEngine { get; }
+
+    public override int Run(EmptyCommandOptions options, InvocationContext context)
+    {
+        ActivityEngine.Dispatch(new LoadServiceProviderActivity());
+        ActivityEngine.Wait();
+        return 0;
+    }
+}
```
mscottford, line 26: I really like using this to test that services are getting loaded correctly. Nicely done!

Corgibytes.Freshli.Cli/CommandRunners/AnalyzeRunner.cs
```diff
@@ -0,0 +1,65 @@
...
+        });
+
+        _activityEngine.Wait();
+
+        return 0;
```
mscottford, line 63: There is no error checking going on in this method. Should there be? I would also expect us to return a non-zero result if the analysis did not correctly run to completion.

Corgibytes.Freshli.Cli/Functionality/Git/CloneGitRepositoryActivity.cs
```diff
@@ -13,45 +13,28 @@ public class CloneGitRepositoryActivity : IApplicationActivity
...
-
     public void Handle(IApplicationEventEngine eventClient)
     {
         // Clone or pull the given repository and branch.
         try
         {
-            var gitRepository = _gitSourceRepository.CloneOrPull(_repoUrl, _branch, _cacheDir, _gitPath);
+            // Null warning suppressed, if we can't fetch it from the service container we got way bigger problems.
+            var gitRepository =
+                eventClient.ServiceProvider.GetService<ICachedGitSourceRepository>()!.CloneOrPull(_repoUrl, _branch,
```
mscottford, line 36: You can use `GetRequiredService` to get rid of the `!`.

features/analyze.feature
```diff
@@ -0,0 +1,53 @@
...
+    The primary user-facing command. This command will delegate to other freshli activities to accomplish its work. It will manage work queues to enable parallelization.
+
+    ```
+    freshli [global options] analyze [command options] <repo url>|<local dir>
+    ```
+
+    With no options specified, performs analysis locally, and then sends the results to the Freshli web app so that the results can be viewed at at URL that will be provided in the command output.
+
+
+    NOTE: These scenarios are not complete nor properly tested because at the time of writing not everything is done. So the entire chain can't be tested yet.
```
mscottford, line 11: What hasn't been done yet? We'll need to create issues for them, and also create an issue to ensure that these scenarios are fleshed out.

mscottford: Also... why wouldn't it make sense to complete everything as part of this pull request?

mscottford (dismissed):
Lots of work here! I have some comments and questions. Some minor. Some will likely lead to the creation of other issues.

edwinkortman:
> Let's move controlling the worker count via a parameter to a separate issue, since there were challenges getting it to work.

Added #232 to iteration 9 but as MVP nice-to-have. 


> Are #205 and #208 also fixed by this pull request? If so, we should correctly note that in the description.
> From reading through the changes, it looks like this pull request also addresses #213. If that's true, please also add a "Fixes..." line for it to the description of this pull request.

Nope, neither of those issues are fixed in this PR. I have made some changes to `CloneGitRepositoryActivity` but the work for those issues still need to be done.

mscottford:
> > Let's move controlling the worker count via a parameter to a separate issue, since there were challenges getting it to work.
> 
> Added #232 to iteration 9 but as MVP nice-to-have.

Yay!
 
> > Are #205 and #208 also fixed by this pull request? If so, we should correctly note that in the description.
> > From reading through the changes, it looks like this pull request also addresses #213. If that's true, please also add a "Fixes..." line for it to the description of this pull request.
> 
> Nope, neither of those issues are fixed in this PR. I have made some changes to `CloneGitRepositoryActivity` but the work for those issues still need to be done.

Assuming that this pull request gets merged in first, we'll need to strengthen the description of those issues so that acceptance tests get properly added to the `analyze` command along the way.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
edwinkortman: That's definitely not correct. I didn't realize that. Thanks! 
The only value that is nullable is branch. https://github.com/corgibytes/freshli-cli/pull/212/commits/cb9a824809fa09908059b1779e63e790c6262be0 this commit reflects that.

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
edwinkortman: Addressed in https://github.com/corgibytes/freshli-cli/pull/212/commits/cb9a824809fa09908059b1779e63e790c6262be0

Corgibytes.Freshli.Cli.Test/CommandOptions/AnalyzeCommandOptionsTest.cs
edwinkortman: Addressed in https://github.com/corgibytes/freshli-cli/pull/212/commits/cb9a824809fa09908059b1779e63e790c6262be0

Corgibytes.Freshli.Cli/CommandRunners/AnalyzeRunner.cs
edwinkortman: Haha, yeah I had the same conclusion but I didn't want to add it in this PR. Same with the acceptance tests, I wanted to wait until everything is connected before finalizing them as the tests depend on this output. 

I like the idea of logging it to a log instead of the console output. Maybe add a `--verbose` option to see these sort of logs?

Corgibytes.Freshli.Cli/CommandRunners/ComputeLibYearCommandRunner.cs
edwinkortman: Yup! I totally forgot about that.

Corgibytes.Freshli.Cli/CommandRunners/Git/GitCloneCommandRunner.cs
```diff
@@ -15,22 +15,19 @@ public class GitCloneCommandRunner : CommandRunner<GitCloneCommand, GitCloneComm
...
-        _gitSourceRepository = gitSourceRepository;
         _activityEngine = activityEngine;
         _eventEngine = eventEngine;
     }
 
     public override int Run(GitCloneCommandOptions options, InvocationContext context)
     {
-        _activityEngine.Dispatch(new CloneGitRepositoryActivity(_gitSourceRepository,
-            options.RepoUrl, options.Branch, options.CacheDir, options.GitPath));
+        // RepositoryUrl, Branch, CacheDir, GitPath, AnalysisId
```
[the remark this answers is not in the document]
edwinkortman: Whoops, that's what I sometimes use to get the parameters in the correct order so I don't have to check the method signature each time.

Corgibytes.Freshli.Cli/CommandRunners/LoadServiceCommandRunner.cs
edwinkortman: Thank you! :D

features/analyze.feature
edwinkortman: I'll make sure to create the issue for it. 
Why it doesn't make sense to complete everything is that it depends on some other issues that aren't done yet:
https://github.com/corgibytes/freshli-cli/issues/205
https://github.com/corgibytes/freshli-cli/issues/208

I'm going to check the entire flow and see if we are missing any connections and make issues for it. After that's done we can update these features.

Corgibytes.Freshli.Cli/Functionality/FreshliWeb/ResultsApi.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+
+namespace Corgibytes.Freshli.Cli.Functionality.FreshliWeb;
+
+public class ResultsApi : IResultsApi
+{
+    public string GetResultsUrl(Guid analysisId) =>
+        // Not the final implementation
+        "https://freshli.app/" + analysisId;
```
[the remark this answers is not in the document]
edwinkortman: Haha I know right.

Corgibytes.Freshli.Cli/CommandRunners/AnalyzeRunner.cs
edwinkortman: There should be, I guess we have to build something that can listen to any error event (instead of having to listen to each error event individually). I'll add an issue for it.

features/analyze.feature
edwinkortman: I've added issues and linked them to the epic here: 
https://github.com/corgibytes/freshli-cli/issues/135#issuecomment-1241804496
There's also an updated image of the happy path flow to show what we are missing because I lost the overview entirely and misguided Dona twice now.
https://github.com/corgibytes/freshli-cli/issues/135#issuecomment-1241805801

Corgibytes.Freshli.Cli/CommandRunners/AnalyzeRunner.cs
edwinkortman: Continuing in issue: https://github.com/corgibytes/freshli-cli/issues/240

edwinkortman: Also related: https://github.com/corgibytes/freshli-cli/issues/241

edwinkortman:
> Assuming that this pull request gets merged in first, we'll need to strengthen the description of those issues so that acceptance tests get properly added to the analyze command along the way.

Yes, I agree but the feature test is depending on the entire chain to work before the output can be properly tested. 
Added an issue for this: https://github.com/corgibytes/freshli-cli/issues/240

PR #216, opened 2022-09-01
Feature/git clone repository

### LINKED ISSUE #215: Let AnalysisStartedEvent.Handle pass CloneGitRepositoryActivity to the activity engine
Blocked by https://github.com/corgibytes/freshli-cli/issues/148
This comes from https://github.com/corgibytes/freshli-cli/issues/148 but is also dependent on https://github.com/corgibytes/freshli-cli/issues/148 finished & merged first because that's the one adding the Activity.

- [ ] AnalysisStartedEvent.Handle
    - [ ] Passes the CloneGitRepositoryActivity to the message bus

klhoot:
Fixes #215 

This change adds an interface that uses a saved analysis to determine the URL and branch to clone from in the CloneGitRepositoryActivity.  The saved analysis ID is then passed to the next stage of the pipeline from the GitRepositoryClonedEvent.

The cache directory and git path are still required to be passed to the CloneGitRepositoryActivity in order to find the saved analysis in the cacheDB and to find the git executable in order to execute the git command.

### DIFF
--- a/Corgibytes.Freshli.Cli/Functionality/Git/CloneGitRepositoryActivity.cs
+++ b/Corgibytes.Freshli.Cli/Functionality/Git/CloneGitRepositoryActivity.cs
@@ -1,25 +1,48 @@
+using System;
 using Corgibytes.Freshli.Cli.Exceptions;
 using Corgibytes.Freshli.Cli.Functionality.Engine;
+using Microsoft.Extensions.DependencyInjection;
 using Newtonsoft.Json;
 
 namespace Corgibytes.Freshli.Cli.Functionality.Git;
 
 public class CloneGitRepositoryActivity : IApplicationActivity
 {
-    [JsonProperty] private readonly string? _branch;
+    [JsonProperty] private readonly ICachedGitSourceRepository _gitSourceRepository;
+
     [JsonProperty] private readonly string _cacheDir;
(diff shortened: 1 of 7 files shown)


Corgibytes.Freshli.Cli/Functionality/Git/CloneGitRepositoryActivity.cs
```diff
@@ -1,25 +1,48 @@
...
+        _analysisId = analysisId;
+    }
+
+    public CloneGitRepositoryActivity(IServiceProvider serviceProvider, Guid analysisId, string cacheDir, string gitPath)
+    {
+        _cacheDir = cacheDir;
+        _gitPath = gitPath;
+        _analysisId = analysisId;
+
+        _gitSourceRepository = serviceProvider.GetRequiredService<ICachedGitSourceRepository>();
```
mscottford, line 38: I've been contemplating whether or not to take this approach with some of the activities. I like seeing it here so we can compare and contrast it against the other approaches and decide on the one that we like most.

mscottford (approved):
This looks good. Nice and clean and straightforward.

I suspect that some changes might still be needed when we go to stitch everything together via the `analyze` command. But let's wait on trying to figure that out for now.

PR #224, opened 2022-09-07
Service provider always null when used in Activities

edwinkortman:
I've tested a few things before ending up on this solution. 

## `JsonContractResolver`

What I started with was @mscottford suggestion to use the `JsonContractResolver`:
```csharp
...
        else if (objectType == typeof(IServiceProvider))
        {
            var serviceProvider = _serviceScopeFactory.CreateScope().ServiceProvider;

            if (serviceProvider != null)
            {
                baseContract.DefaultCreator = () => serviceProvider;
            }
        }
...
```
This results in a self referencing loop, though. Potential solutions online for this problem are either: ignore it, ignore it globally or remove it. Ignoring it with `[JsonIgnore]` does make the loop error go away, but it also ignores the contract resolver.

Examples:
https://makolyte.com/newtonsoft-json-jsonserializationexception-self-referencing-loop-detected-for-property/

## Pass the service collection into the handle method

```csharp
public interface IApplicationActivity
{
    public void Handle(IApplicationEventEngine eventClient, IServiceProvider);
}
```

So, this does work. One of the concerns was that it might not work because it also gets serialized. Running the test added in https://github.com/corgibytes/freshli-cli/commit/d64cbdd26b274a1c56a933166c4657bbd3279283 did not show any problems with serializing. A concern I do have is that the objects enqueued will be huge with the entire service provider added unless there's some hidden things going on that reduces it. I did not look into this because I had an epiphany. 

## Solution proposed

The epiphany I had was when step debugging the previous solution with adding it to the Handle is that I had the ServiceProvider available in the ApplicationEngine. Instead of going the long way, I've made it a public property so from an Activity you are able to call the Provider directly. This way it doesn't need to be injected and it doesn't get serialized. 

I think long-term we should still move towards splitting up the Activities from their Handlers. What this PR proposes is basically using the container as a Service Locator which has been described as an anti-pattern by some. I personally prefer to have dependency injection as well as testing things with service locators feel cumbersome. But I also know I'm biased.

https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/
> Let's examine why this is so. In short, the problem with Service Locator is that it hides a class' dependencies, causing run-time errors instead of compile-time errors, as well as making the code more difficult to maintain because it becomes unclear when you would be introducing a breaking change.

### DIFF
--- a/Corgibytes.Freshli.Cli.Test/ProgramTest.cs
+++ b/Corgibytes.Freshli.Cli.Test/ProgramTest.cs
@@ -75,4 +75,15 @@ public class ProgramTest : FreshliTest
             "^ERROR|.*System.Exception: Simulating failure from an activity$", RegexOptions.Multiline
         ));
     }
+
+    [Fact]
+    public void ValidateServiceProviderIsLoaded()
+    {
+        MainCommand.ShouldIncludeLoadServiceCommand = true;
+        var task = Task.Run(() => Program.Main("load-service"));
+        task.Wait();
+
+        _consoleOutput.ToString().Should().Contain("All good! Service provider is not null.");
+        _consoleOutput.ToString().Should().NotContain("Simulating loading the service provider, but the provider is null.");
+    }
 }
--- a/Corgibytes.Freshli.Cli/Functionality/Analysis/CacheWasNotPreparedEvent.cs
+++ b/Corgibytes.Freshli.Cli/Functionality/Analysis/CacheWasNotPreparedEvent.cs
@@ -1,4 +1,3 @@
-using System;
 using Corgibytes.Freshli.Cli.CommandRunners.Cache;
 using Corgibytes.Freshli.Cli.Functionality.Engine;
 using Microsoft.Extensions.DependencyInjection;
@@ -7,8 +6,6 @@ namespace Corgibytes.Freshli.Cli.Functionality.Analysis;
 
 public class CacheWasNotPreparedEvent : ErrorEvent
 {
-    public IServiceProvider ServiceProvider { get; init; } = null!;
-
     public string GitPath { get; init; } = null!;
     public string CacheDirectory { get; init; } = null!;
     public string RepositoryUrl { get; init; } = null!;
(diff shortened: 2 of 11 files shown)


Corgibytes.Freshli.Cli.Test/ProgramTest.cs
```diff
@@ -75,4 +75,15 @@ public void ValidateExceptionsInActivityHandlersWriteToLog()
...
+    [Fact]
+    public void ValidateServiceProviderIsLoaded()
+    {
+        MainCommand.ShouldIncludeLoadServiceCommand = true;
+        var task = Task.Run(() => Program.Main("load-service"));
+        task.Wait();
+
+        _consoleOutput.ToString().Should().Contain("All good! Service provider is not null.");
+        _consoleOutput.ToString().Should().NotContain("Simulating loading the service provider, but the provider is null.");
+    }
```
mscottford, line 88: I like this! Nicely done.

Corgibytes.Freshli.Cli/Functionality/Analysis/CacheWasNotPreparedEvent.cs
```diff
@@ -26,8 +23,8 @@ public override void Handle(IApplicationActivityEngine eventClient)
         });
 
         eventClient.Dispatch(new RestartAnalysisActivity(
-            ServiceProvider.GetRequiredService<ICacheManager>(),
-            ServiceProvider.GetRequiredService<IHistoryIntervalParser>()
+            eventClient.ServiceProvider.GetRequiredService<ICacheManager>(),
+            eventClient.ServiceProvider.GetRequiredService<IHistoryIntervalParser>()
```
mscottford, line 27: I like this compromise.

mscottford (approved):
I think this is a good compromise to resolve the issue at hand. It lets us move forward without too much trouble. I agree that we should add split-up Activity/Event classes from handling logic, and I also feel that it needs to wait for a little bit.

Corgibytes.Freshli.Cli.Test/ProgramTest.cs
edwinkortman: Thank you! :D

PR #243, opened 2022-09-09
Compute history intervals

edwinkortman:
Fixes #213

PR #250, opened 2022-09-13
Compute history intervals

### LINKED ISSUE #213: Compute history should be able to work with custom intervals like 2weeks or 3months
Currently it can't. Currently the service is only able to do 1 year, 1 month, 1 week or 1 day intervals. One requirement in [analyze](https://github.com/corgibytes/freshli-cli/issues/56) is to make it work with a custom interval.

- default value: 1m
- valid values: /\d+[ymwd]?/
    - y suffix – specifies duration in years
    - m suffix – specifies duration in months
    - w suffix – specifies duration in weeks
    - d suffix – specifies duration in days

This, at least, needs to be refactored in:
- `Corgibytes.Freshli.Cli/Functionality/Git/ComputeHistory.cs`

edwinkortman:
Fixes #213 
I accidentally merged the wrong PR: https://github.com/corgibytes/freshli-cli/pull/243
That PR has been reverted in: https://github.com/corgibytes/freshli-cli/pull/249

Related files in lib:
https://github.com/corgibytes/freshli-lib/blob/main/Corgibytes.Freshli.Lib.Test/Unit/AnalysisDatesTest.cs
https://github.com/corgibytes/freshli-lib/blob/main/Corgibytes.Freshli.Lib/AnalysisDates.cs


Corgibytes.Freshli.Cli/Functionality/Git/ComputeHistory.cs
```diff
@@ -9,48 +8,109 @@ namespace Corgibytes.Freshli.Cli.Functionality.Git;
...
+                rangeStartDate = new DateTimeOffset(rangeStartDate.Year, rangeStartDate.Month, 1, 0, 0, 0, rangeStartDate.Offset);
+                break;
+            case "y":
+                // Start at first day of the year
+                rangeStartDate = new DateTimeOffset(rangeStartDate.Year, 1, 1, 0, 0, 0, rangeStartDate.Offset);
+                break;
+        }
+
+        return rangeStartDate;
     }
```
klhoot, line 114: I am not understanding why the range start date is being forced to a Monday, the first day of the Month, or the first day of the year.  I was thinking that the interval parser was going to handle arbitrary date ranges.

mscottford: This logic is inherited from `freshli-lib`. You start on the current day, go to the start of the interval that the current day is in the middle of, and then skip backwards in time every X (days, weeks, months, or years depending on the selection). `freshli-lib` only supported an interval jump of 1 month, but this is the way that it handled it. This was needed for the month logic because a month is not a fixed length of time (not every month has a 31st day, but every month has a 1st day). In addition to the skips, the date of the oldest commit is always included, and the current date is always included. This means that the interval between the first and last pairs of data is not the same as the interval between the rest of the pairs of data.

Corgibytes.Freshli.Cli.Test/Functionality/Git/ComputeHistoryTest.cs
```diff
@@ -14,167 +16,202 @@ namespace Corgibytes.Freshli.Cli.Test.Functionality.Git;
...
+                    new("ef14791d014431952aa721fa2a9b22afb8d4f144",
+                        new DateTimeOffset(2021, 1, 3, 00, 00, 00, TimeSpan.Zero)),
+                    new("4f6b7990ad45b2c5bf5817c359de72729654dd9f",
+                        new DateTimeOffset(2021, 1, 2, 00, 00, 00, TimeSpan.Zero)),
+                    new("4f6b7990ad45b2c5bf5817c359de72729654dd9f",
+                        new DateTimeOffset(2021, 1, 1, 00, 00, 00, TimeSpan.Zero)),
+                    // End date
+                    new("4f6b7990ad45b2c5bf5817c359de72729654dd9f",
+                        new DateTimeOffset(2020, 12, 31, 00, 00, 00, TimeSpan.Zero))
+                }
```
mscottford, line 130: Yay! This looks good.

Corgibytes.Freshli.Cli/Functionality/Git/ComputeHistory.cs
```diff
@@ -9,48 +8,109 @@ namespace Corgibytes.Freshli.Cli.Functionality.Git;
...
-                from groupedHistory in groupedHistories
-                select groupedHistory.MaxBy(i => i.CommittedAt)
-                into mostRecentCommitForThisGroupedPeriod
-                where mostRecentCommitForThisGroupedPeriod != null
-                select new HistoryIntervalStop(mostRecentCommitForThisGroupedPeriod.ShaIdentifier,
-                    mostRecentCommitForThisGroupedPeriod.CommittedAt))
-            .ToList();
+            from offset in range
+            let lastCommitForOffset = gitCommits.First(commit => commit.CommittedAt <= offset)
+            select new HistoryIntervalStop(lastCommitForOffset.ShaIdentifier, offset))
```
mscottford, line 87: I'm not a fan of this style of LINQ expressions. Can we please switch to the other style? I wish I knew a way to make a linter rule for this. :)

mscottford: Actually, now that I think about it. I'm okay with this for now. I can create an issue for us to add a linter rule for this instead.

mscottford (approved):

PR #207, opened 2022-08-30
Bill of materials

edwinkortman:
Fixes #152 

This is not done yet, adding another issue for it as it's depending on this to be merged first: https://github.com/corgibytes/freshli-cli/pull/204/files#diff-8b3691b75312db3758ae2ac096f46f86676c6534002d0cb8832fd462933b4de0
 - [ ] HistoryIntervalStopFoundEvent
     - [ ] Passes the message GenerateBillOfMaterialsActivity to the message bus

PR #275, opened 2022-09-19
Let HistoryStopCheckedOutEvent.Handle pass DetectAgentsForDetectManif…

### LINKED ISSUE #237: Let `HistoryStopCheckedOutEvent.Handle` pass `DetectAgentsForDetectManifestsActivity` to the Activity Engine
- [x] `HistoryStopCheckedOutEvent.Handle`
    - [x] `DetectAgentsForDetectManifestsActivity`

donabp:
Fixes #237

This spawns from #207


mscottford (approved):
I'd ideally like to see some tests for this change, but it's small enough that I think we can proceed with it as is.

PR #282, opened 2022-09-21
Fixes for compute history and checkout history activities

klhoot:
Fixes deserialization of CheckoutHistoryActivity - because analysisLocation was being used to initialize _cacheDirectory, _repositoryId, and _commitId, but was not saved in the object, the values were not being deserialized correctly when the activity ran.  This change saves _analysisLocation in the object instead and uses _analysisLocation directly in Handle.  This change also passes _analysisLocation to HistoryStopCheckedOutEvent instead of creating a new AnalysisLocation instance.

Fixes HistoryIntervalStopFoundEvent which was expecting the Git commit Id to be provided in AnalysisLocation and was not using the GitCommitIdentifier field. This change also creates a new historyStopLocation and passes that to the event instead of overwriting the values in the same AnalysisLocation which *may* work depending on when the object is serialized, but feels unnecessarily risky.


mscottford (approved):
This looks great!

PR #284, opened 2022-09-21
Returns and empty list of detect-manifests returns an empty string

klhoot:
Ensures that the list of manifest files is empty, as expected when no manifest files are found by the agent.

### DIFF
--- a/Corgibytes.Freshli.Cli.Test/Services/AgentReaderTest.cs
+++ b/Corgibytes.Freshli.Cli.Test/Services/AgentReaderTest.cs
@@ -44,8 +44,31 @@ public class AgentReaderTest
         checkoutDirectory.Delete(true);
     }
 
+    [Fact]
+    public void AgentReaderReturnsEmptyListWhenNoManifestsFound()
+    {
+        var checkoutLocation = CreateCheckoutLocation(out var checkoutDirectory);
+        var reader = new AgentReader("freshli-agent-java");
+        var repositoryLocation = Path.Combine(checkoutLocation, "invalid_repository");
+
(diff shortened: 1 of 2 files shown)


Corgibytes.Freshli.Cli/Services/AgentReader.cs
```diff
@@ -47,7 +48,8 @@ public List<string> DetectManifests(string projectPath)
     {
         var manifests = Invoke.Command(AgentExecutablePath, $"detect-manifests {projectPath}", ".");
 
-        return manifests.TrimEnd('\n', '\r').Split("\n").ToList();
+        return manifests.IsEmpty() ? new List<string>() :
```
edwinkortman, line 51: Could we add a test for this?

klhoot: Good suggestion!

Corgibytes.Freshli.Cli.Test/Services/AgentReaderTest.cs
```diff
@@ -44,8 +44,31 @@ public void GenerateBillOfMaterialsUsingProtobuf()
         checkoutDirectory.Delete(true);
     }
 
+    [Fact]
+    public void AgentReaderReturnsEmptyListWhenNoManifestsFound()
+    {
+        var checkoutLocation = CreateCheckoutLocation(out var checkoutDirectory);
```
edwinkortman, line 50:
```suggestion
        var checkoutLocation = CreateCheckoutLocation(out checkoutDirectory);
```

edwinkortman (approved):
Nice work! :D

PR #289, opened 2022-09-23
Agent bugs

edwinkortman:
Found two small bugs. One is that I forgot a DateTime format, the other one was that the working directory wasn't passed so we were asking the agent to run this:
```shell
/usr/local/bin/freshli-agent-java process-manifest pom.xml 2022-09-23T13:37:24Z
```

Which doesn't work because it can't locate `pom.xml`


edwinkortman:
Ignore my request to review, the tests are all green but actually using it with working directory doesn't work.

mscottford (approved):
Looks good to me.

PR #290, opened 2022-09-23
Add latest only option

### LINKED ISSUE #234: Add option to `analyze` command so that it does not collect historical metrics
The `--latest-only` option changes the behavior of the latest command so that it doesn't walk the git history for the provided repository and branch. This option is incompatible with the `--include-history` and `--history-interval` options.

We should also permit analyzing a codebase that has already been checked out (should that be a separate issue?). This is done by providing a path, such as `~/code/project` instead of a url for a Git repository. When analyzing a local directory, if `--latest-only` is provided, then it is permissible for the local directory to contain changed files. Otherwise, a clean working directory is required.

We should also permit appending the data collected to a previous set of analysis results with the `--analysis-results-id` option. Make this option only allowed when `--latest-only` is specified.

edwinkortman:
Fixes #234

This has https://github.com/corgibytes/freshli-cli/pull/289 already merged into it because it was blocking me from testing it properly. :-)


Corgibytes.Freshli.Cli/Functionality/LatestOnly.cs
```diff
@@ -0,0 +1,7 @@
+namespace Corgibytes.Freshli.Cli.Functionality;
+
+public enum LatestOnly
+{
+    ShowLatestOnly,
+    WalkBackInRevisionHistory
+}
```
mscottford, line 7: I feel like these names could be improved upon. What do you think of:
```suggestion
public enum RevisionHistoryMode
{
    OnlyLatestRevision,
    AllRevisions
}
```

mscottford: After writing that, I think that it's an improvement, and I suspect we could still do better than that. :)

mscottford (approved):
I'm marking this as approved even though I think we could come up with a better name. If you agree that a different name is worth the effort, then go ahead and change it before merging. If you like the name the way it is, then merge it in as is.

Corgibytes.Freshli.Cli/Functionality/LatestOnly.cs
edwinkortman: Ah yes! Much better, thank you.

edwinkortman: I've used Rider to do the rename though :D

PR #291, opened 2022-09-26
Introduces IConfiguration interface

mscottford:
The goal of this interface is to provide storage for configuration values that are needed throughout the application. For starters these are the path to the application cache and the path to the git executable, both of which can be specified via parameters when running the `analyze` command. Using this configuration interface means that we don't need to pass these values around as much, and instead can pull them out of the configuration when they are needed.

I also took the time to refactor the activities to pull services from the `IServiceProvider` instead of having them passed in as constructor values.


Corgibytes.Freshli.Cli/Functionality/Configuration.cs
```diff
@@ -0,0 +1,16 @@
...
+public class Configuration : IConfiguration
+{
+    public string GitPath { get; set; }
+    public string CacheDir { get; set; }
+
+    public Configuration(IEnvironment environment)
+    {
+        GitPath = "git";
+        CacheDir = Path.Combine(environment.HomeDirectory, ".freshli");
+    }
```
edwinkortman, line 11: This means we have default values in two places. We probably want to use these values in `AnalyzeCommand` as well.

edwinkortman (approved):
Great change! Thanks. We do have "git" and the cache dir set as default value in two places now though. If that's addressed, the rest is okay for me! :+1:

Corgibytes.Freshli.Cli/Functionality/Configuration.cs
mscottford: That's a good point about default values being in two places. I'll take a look at that.

mscottford: This should be addressed in 6984249.

mscottford: I found another one. Fixed in 573f68b.

PR #293, opened 2022-09-26
Fixes deserialization of Configuration class in Newtonsoft.json

klhoot:
For the configuration object to be deserialized correctly in AnalysisLocation there needs to be a constructor that will set GitPath and CacheDir from the serialized data.  Because environment is not saved in the class it cannot be used to deserialize the object.


edwinkortman:
Very nice! Thanks @klhoot

PR #295, opened 2022-09-27
Update acceptance tests for analyze command

### LINKED ISSUE #240: Flesh out acceptance tests
Spawned from: https://github.com/corgibytes/freshli-cli/pull/212#discussion_r965984267, https://github.com/corgibytes/freshli-cli/pull/212#discussion_r966005383

The scenarios in `analyze.feature` are currently a little meager. 

### This issue is blocked until the entire happy path has been implemented. Blocked by:
- https://github.com/corgibytes/freshli-cli/issues/205
- https://github.com/corgibytes/freshli-cli/issues/208
- https://github.com/corgibytes/freshli-cli/issues/237
- https://github.com/corgibytes/freshli-cli/issues/238
- https://github.com/corgibytes/freshli-cli/issues/239

### To do
- [ ] Test: Run the analysis with default options, `freshli analyze https://github.com/corgibytes/freshli-fixture-ruby-nokotest`. Running it with just the repository passed, should lead to using all the default options. 
- [ ] Test: Run the analysis with specific git installation, `freshli analyze --git-path=git https://github.com/corgibytes/freshli-fixture-ruby-nokotest`
- [ ] Test: Run the analysis for a specific branch. e.g. `freshli analyze --branch=trunk https://github.com/corgibytes/freshli-fixture-ruby-nokotest`
- [ ] Test: Run the analysis for every point in time when the files have changed. `freshli analyze --commit-history https://github.com/corgibytes/freshli-fixture-ruby-nokotest`
- [ ] Test: Run the analysis at a specific interval. In the example: take the last point in time per three months that files have changed. `freshli analyze --history-interval=3m https://github.com/corgibytes/freshli-fixture-ruby-nokotest`
- [ ] Test: Run the analysis, start with 6 workers. `freshli analyze --workers=6 https://github.com/corgibytes/freshli-fixture-ruby-nokotest`

klhoot:
Fixes #240.


features/analyze.feature
```diff
@@ -8,60 +8,130 @@ Feature: analyze
     With no options specified, performs analysis locally, and then sends the results to the Freshli web app so that the results can be viewed at at URL that will be provided in the command output.
 
 
-    NOTE: These scenarios are not complete nor properly tested because at the time of writing not everything is done. So the entire chain can't be tested yet.
+    NOTE: These scenarios are not complete the Web API is not implemented yet. So the results can't be verified yet.
```
mscottford, line 11: I would have thought that the contents of the cache database could be used to determine if the analysis was completed correctly or contained the correct results. The current version of the web API is essentially write-only, so it's not possible at this point in time to use it to validate that data collection has happened correctly.

mscottford (commented):
It's nice to see the return of the validation logic that we had for the commands that have been removed. I was expecting to see some validation of the actual history stops and the LibYear values for those stops. I don't think we need an exhaustive check, but a few spot checks would go a long way, and I think we can pull data out of the cache database to determine that.

features/analyze.feature
klhoot: I want to have checks for the history stops, etc but I also wanted to raise the pull request to show some progress now that I have gotten past the bugs that were found while updating the tests.
I started working to figure out where/how I could get access to that information from an acceptance test - I was tracking through the code and thinking about console output, but had not thought to look in the database so I will try that! Thanks

mscottford:
Merging after review and cleanup during ensemble session.

PR #299, opened 2022-09-30
Analyze codebase already checked out

### LINKED ISSUE #287: Analyze a codebase that already has been checked out

edwinkortman:
Fixes #287


Corgibytes.Freshli.Cli/Functionality/Analysis/AnalysisLocation.cs
```diff
@@ -13,13 +13,22 @@ public AnalysisLocation(IConfiguration configuration, string repositoryId, strin
...
     {
         get
         {
+            if (LocalDirectory != null)
+            {
+                return CommitId == null
+                    ? LocalDirectory
+                    : System.IO.Path.Combine(_configuration.CacheDir, "histories", RepositoryId, CommitId);
+            }
+
```
mscottford, line 31: Is there a way to implement this that doesn't duplicate the logic on line 37?

Would this work?

```csharp
if (LocalDirectory != null)
{
    return LocalDirectory;
}
```

And then the rest of the method as it was before?

Corgibytes.Freshli.Cli/Functionality/Git/CachedGitSourceRepository.cs
```diff
@@ -32,21 +30,20 @@ public CachedGitSource FindOneByHash(string hash)
...
-        using var sha256 = SHA256.Create();
-        var hashBytes = sha256.ComputeHash(Encoding.UTF8.GetBytes(url + branch));
-        var stringBuilder = new StringBuilder();
-        foreach (var hashByte in hashBytes)
-        {
-            stringBuilder.Append(hashByte.ToString("x2"));
-        }
-
-        var hash = stringBuilder.ToString();
+        var hash = new CachedGitSourceId(url, branch).Id;
```
mscottford, line 46: I like having this logic live in a class. Nice refactoring!

Corgibytes.Freshli.Cli/Functionality/Git/GitManager.cs
```diff
@@ -6,14 +7,24 @@ public class GitManager : IGitManager
...
-    {
-        _gitArchive = gitArchive;
-    }
+    public GitManager(GitArchive gitArchive) => _gitArchive = gitArchive;
 
     public string CreateArchive(
         string repositoryId, GitCommitIdentifier gitCommitIdentifier) =>
         _gitArchive.CreateArchive(repositoryId, gitCommitIdentifier);
 
+    public bool GitRepositoryInitialized(string repositoryLocation, IConfiguration configuration)
```
mscottford, line 16: I think `IConfiguration` is better injected via the constructor then forcing the caller to know how to pass it in.

Corgibytes.Freshli.Cli/Functionality/Git/IGitManager.cs
```diff
@@ -5,4 +5,6 @@ public interface IGitManager
     public GitCommitIdentifier ParseCommitId(string commitId);
 
     string CreateArchive(string repositoryId, GitCommitIdentifier gitCommitIdentifier);
+
+    public bool GitRepositoryInitialized(string repositoryLocation, IConfiguration configuration);
```
mscottford, line 9: I think this name would better indicate that this is a predicate method:

```suggestion
    public bool IsGitRepositoryInitialized(string repositoryLocation, IConfiguration configuration);
```

mscottford (changes requested):
I've got a couple small suggestions.

Corgibytes.Freshli.Cli/Functionality/Git/CachedGitSourceRepository.cs
edwinkortman: Thank you! :D

Corgibytes.Freshli.Cli/Functionality/Analysis/AnalysisLocation.cs
edwinkortman: Not entirely but I did manage it with that check. We should only return the local directory if the commit id is null and the local directory is not null. So it's two if statements in one now but should be fine. :D

```csharp
if (CommitId == null)
{
      if (LocalDirectory != null)
      {
          return LocalDirectory;
      }

      return System.IO.Path.Combine(_configuration.CacheDir, "repositories", RepositoryId);
}

return System.IO.Path.Combine(_configuration.CacheDir, "histories", RepositoryId, CommitId);
```

edwinkortman:
Thank you for the review @mscottford :D

PR #307, opened 2022-10-03
Adds CachedLibYears table to the database

klhoot:


mscottford:
Merging this in as is, even though there are failures. Some of the fixes in here will benefit other investigations.

PR #310, opened 2022-10-04
Start x workers

### LINKED ISSUE #232: Adjust number of workers that should be running
Spawned from https://github.com/corgibytes/freshli-cli/pull/212#issuecomment-1240704175 and https://github.com/corgibytes/freshli-cli/issues/56.

The number of worker processes that should be running at any given time. This defaults to twice the number of CPU cores.

edwinkortman:
Fixes #232 

I decided to move the `--workers` from the `AnalyzeCommand` to the `MainCommand` because otherwise this check will fail when running a command that's not the analyze command:
```csharp
var workerCount = context.ParseResult.GetOptionValueByName<int>("workers");
```

It fails because the workers option doesn't exist at all. An alternative would be to catch the exception I guess but I found this way nicer. Maybe there's another way I'm missing? 

The statistics we have don't contain a worker count so I hope this screenshot will suffice. 
image

### DIFF
--- a/Corgibytes.Freshli.Cli.Test/Commands/MainCommandTest.cs
+++ b/Corgibytes.Freshli.Cli.Test/Commands/MainCommandTest.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.CommandLine;
 using Corgibytes.Freshli.Cli.Commands;
 using Corgibytes.Freshli.Cli.Functionality;
@@ -25,11 +26,24 @@ public class MainCommandTest : FreshliTest
     }
 
     [Theory]
-    [InlineData("--cache-dir")]
-    public void Verify_cache_dir_option_configuration(string alias)
+    [MemberData(nameof(OptionsData))]
+    public void VerifyOptions(string alias, ArgumentArity arity)
     {
         var configuration = new Mock<IConfiguration>();
         var mainCommand = new MainCommand(configuration.Object);
-        mainCommand.VerifyAlias(alias, ArgumentArity.ExactlyOne, false);
+        mainCommand.VerifyAlias(alias, arity, false);
     }
+
+    public static IEnumerable<object?[]> OptionsData() =>
+        new List<object?[]>
+        {
+            new object?[]
+            {
+                "--workers", ArgumentArity.ZeroOrOne
+            },
+            new object?[]
+            {
+                "--cache-dir", ArgumentArity.ExactlyOne
+            }
+        };
 }
--- a/Corgibytes.Freshli.Cli/Program.cs
+++ b/Corgibytes.Freshli.Cli/Program.cs
@@ -3,11 +3,13 @@ using System.CommandLine.Builder;
 using System.CommandLine.Hosting;
 using System.CommandLine.Invocation;
 using System.CommandLine.Parsing;
+using System.Threading;
 using System.Threading.Tasks;
 using Corgibytes.Freshli.Cli.Commands;
 using Corgibytes.Freshli.Cli.Extensions;
 using Corgibytes.Freshli.Cli.Functionality;
 using Corgibytes.Freshli.Cli.IoC;
+using Hangfire;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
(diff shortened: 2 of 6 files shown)


Corgibytes.Freshli.Cli/Program.cs
```diff
@@ -91,6 +97,19 @@ public static CommandLineBuilder CreateCommandLineBuilder()
...
+            .AddMiddleware(async (context, next) =>
+            {
+                var workerCount = context.ParseResult.GetOptionValueByName<int>("workers");
+                HangfireBackgroundService?.StopAsync(CancellationToken.None);
+                if (HangfireOptions != null && workerCount > 0)
+                {
+                    HangfireOptions.WorkerCount = workerCount;
+                }
+
+                HangfireBackgroundService?.StartAsync(new CancellationToken());
```
mscottford, line 109: Oh! This reminds me of an issue that we need to create. If I remember correctly, we're not able to stop the program with `Ctrl-C` once it's running. We should have a way to "listen" for that keypress and then signal the `CancelationToken` when it happens.

Corgibytes.Freshli.Cli/Program.cs
```diff
@@ -91,6 +97,19 @@ public static CommandLineBuilder CreateCommandLineBuilder()
                 await next(context);
             })
             .AddMiddleware(async (context, next) => { await LogExecution(context, next); })
+            .AddMiddleware(async (context, next) =>
+            {
+                var workerCount = context.ParseResult.GetOptionValueByName<int>("workers");
+                HangfireBackgroundService?.StopAsync(CancellationToken.None);
```
mscottford, line 103: Instead of stopping the service here, should we instead just not start it in the `FreshliServiceBuilder`?

If we decide to keep this stop call here, we should make this call with `await` so that we don't risk calling `StartAsync` before `StopAsync` has finished.

features/analyze.feature
```diff
@@ -115,7 +115,7 @@ Feature: analyze
 
     Scenario: Run the analysis, start with 6 workers.
         Given a directory named "~/.freshli" does not exist
-        When I run `freshli analyze --workers=6 https://github.com/corgibytes/freshli-fixture-java-test`
+        When I run `freshli --workers=6 analyze https://github.com/corgibytes/freshli-fixture-java-test`
```
mscottford, line 118: I'd love for us to be able to examine the output to determine if 6 workers have actually started. I think that could be possible by increasing the `loglevel` and then examining the command output to see if there are exactly 6 lines that indicate a worker has started.

(The "exactly 6 lines" approach might not work if don't delay starting the `HangfireBackgroundService` until after the middleware runs, however.)

mscottford (commented):
I like moving `--workers` to belong to the `MainCommand`. That makes sense to me since the `HangfireBackgroundService` is a basic building block of how all commands are implemented.

I've got a couple of questions below.

Corgibytes.Freshli.Cli.Test/Commands/MainCommandTest.cs
```diff
@@ -25,11 +26,24 @@ public void Verify_no_handler_configuration()
...
-        mainCommand.VerifyAlias(alias, ArgumentArity.ExactlyOne, false);
+        mainCommand.VerifyAlias(alias, arity, false);
     }
+
+    public static IEnumerable<object?[]> OptionsData() =>
+        new List<object?[]>
+        {
+            new object?[]
+            {
+                "--workers", ArgumentArity.ZeroOrOne
```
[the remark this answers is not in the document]
edwinkortman: You are absolutely right. I've changed it to ExactlyOne. The error you get is that it cannot find the option `--workers`

Corgibytes.Freshli.Cli/Program.cs
edwinkortman: Done #321 :D

Corgibytes.Freshli.Cli/Program.cs
edwinkortman: I'll give the not starting it a try. Adding the `await` shows a memory warning:
image

edwinkortman: Oh without await as well.

features/analyze.feature
edwinkortman: Apparently Hangfire is kind to us and supplies it in one of the first lines as:
```
Worker count: 12
```

edwinkortman:
I'll pick this up again after things have sped up :-) e.g. with some caching https://github.com/corgibytes/freshli-cli/pull/325

PR #311, opened 2022-10-05
Implemented Doctor command

### LINKED ISSUE #231: Add `doctor` command for validating environment assumptions
The `freshli` executable makes several assumptions about the environment that it runs in. Here are just a few:
* The `git` executable is available and runnable
* Each agent executable is available and runnable (this is checked by the `agents verify` command)
* Each agent has the appropriate runtime libraries that it needs to function (this is checked by the `agents verify` command)
* The `cache-dir` is available (or permissions allow creating it) and it's permissible to create files and directories in the `cache-dir`

Having the ability to verify that these assumptions are true would be helpful when attempting to optimize the production Docker image as described in #229 and #230. That is the goal of the command that is described in this issue.

After this command has been built, then container images can be validated by running `freshli doctor && freshli agents verify`.

```
freshli [global options] doctor [command options]
```

**What this command does?**

Validates assumptions about the current execution environment. Specifically:
- [x] It's possible to create the cache directory if it does not exist
- [x] It's possible to create and write to files in the cache directory
- [x] It's possible to create sub-directories in the cache directory
- [x] It's possible to create and write to files in sub-directories in the cache directory
- [x] The `git` executable can be run -- tested by running `git version` and checking/displaying output

The result of running each of these checks is output to the console.

If all of the checks pass, then the command's exit code is 0. If _any_ of the checks fail, then the exit code is not 0.

Global Options

* `--cache-dir <path>`
    * the location where the `freshli` command will write temporary files as part of its processing
    * default value: `$HOME/.freshli`

Command Options

* `--git-path`
    * path to the `git` binary
    * default value: `git` -- when default value is used, it's assumed that the `git` command is accessible without needed a full path

donabp:
Fixes #231


Corgibytes.Freshli.Cli/Commands/DoctorCommand.cs
```diff
@@ -0,0 +1,22 @@
...
+
+public class DoctorCommand : RunnableCommand<DoctorCommand, DoctorCommandOptions>
+{
+    public DoctorCommand() : base("doctor",
+        "Validating environment assumptions.")
+    {
+        var gitPath = new Option<string>(
+            "--git-path",
+            description: "Path to the git binary. Default = 'git'",
+            getDefaultValue: () => "git"
```
edwinkortman, line 14:
```suggestion
            getDefaultValue: () => configuration.GitPath
```

We ideally re-use the configuration object here to give us the default path to git. [AnalyzeCommand](https://github.com/corgibytes/freshli-cli/blob/main/Corgibytes.Freshli.Cli/Commands/AnalyzeCommand.cs#L9) does that for example.

Corgibytes.Freshli.Cli/Functionality/Doctor/StartDoctorActivity.cs
```diff
@@ -0,0 +1,65 @@
...
+        try
+        {
+            if (!Directory.Exists(CacheDirectory))
+            {
+                // It's possible to create the cache directory if it does not exist
+                Directory.CreateDirectory(CacheDirectory);
+                Console.Out.Write("Successfully created the cache directory");
+            }
+
+            File.WriteAllText(CacheDirectory + Path.DirectorySeparatorChar + "TextFile.txt",
```
edwinkortman, line 33: I think we should clean up these files as well.

features/doctor.feature
```diff
@@ -0,0 +1,45 @@
...
+      """
+      Then the output should contain:
+      """
+      Wrote inside the sub directory file successfully
+      """
+      Then the output should contain:
+      """
+      Git version found: git version
+      """
+  Scenario: Validating the assumptions the cache directory exists
```
edwinkortman, line 25: Nice scenarios! It would be good to test out the bad paths as well. So, test what happens when it cannot create the directory? otherwise these bad paths aren't tested somewhere. :-)

edwinkortman (changes requested):

Corgibytes.Freshli.Cli/Functionality/Doctor/StartDoctorActivity.cs
donabp: 👋 Hi @edwinkortman when you are saying clean up these files, do you mean the whitespace?

edwinkortman: The activity is creating files, with cleaning up I mean deleting them after everything has been tested.

donabp: Ah I see, I can get that done, let me finish up the other comments. Thanks for the feedback!

mscottford:
Closing this. It's way too far out of sync now.

Issue #324, opened 2022-10-10
Performance


edwinkortman:

edwinkortman:
## Running `freshli analyze https://github.com/corgibytes/freshli-fixture-java-test` 

![Image](https://user-images.githubusercontent.com/7047894/194848432-bd6dab61-96a2-496f-b2a3-d6f0dd579e10.png)


`Corgibytes.Freshli.Cli.Functionality.Environment.GetListOfFiles` throws 168 exceptions

edwinkortman:
## Adding caching

Caching was added in this PR: https://github.com/corgibytes/freshli-cli/pull/325
If the PR is merged & approved, the Package results are saved to the database saving us from calling the agent multiple times for the same package. 

**Command:** freshli analyze https://github.com/corgibytes/freshli-fixture-java-test
**Total time running:** 1m 21s (before 1m 56s)

edwinkortman:
With these PRs added, I think it's best to revisit this issue once we got those implemented. The main culprits I could identify were calls to the agent.

https://github.com/corgibytes/freshli-agent-java/issues/34
https://github.com/corgibytes/dependency-history-maven/issues/24

PR #325, opened 2022-10-10
Caching

edwinkortman:
Related to #324 

Bit weird but the Package URL is added twice to the database. Once with version, once without. The package url without version allows us to fetch the entire release history for a given package instead of querying it with a LIKE (which would slow it down again). 

## Some statistics
https://github.com/corgibytes/freshli-fixture-java-test
Running `freshli analyze https://github.com/corgibytes/freshli-fixture-java-test`

- `freshli-agent-java retrieve-release-history` called **18** times
- `freshli-agent-java detect-manifests` called **89** times
- `freshli-agent-java process-manifest` called **90** times
- `CreateApiHistoryStopActivity` called **45** times
- `CreateApiPackageLibYearActivity` called **44** times

There's 45 months between now, and the first commit in this repository.


edwinkortman:
Ditching this PR as I didn't notice Kimarie and I were doing the same thing.. :-/
https://github.com/corgibytes/freshli-cli/pull/307

PR #364, opened 2022-10-21
Cache retrieve release history

### LINKED ISSUE #357: Cache calls to `retrieve-release-history`
We're making multiple calls for the same package to each agent's `retrieve-release-history` command. We should save the history information in the cache and then reuse it instead.

mscottford:
Fixes #357

Note: This branch is based on #363. It could probably be rebased off of `main`.


Corgibytes.Freshli.Cli.Test/Services/AgentReaderTest.cs
```diff
@@ -1,85 +1,53 @@
...
+        var agentExecutable = "/path/to/agent";
+        var commandInvoker = new Mock<ICommandInvoker>();
+        var packageUrl = new PackageURL("pkg:maven/org.example/package");
+
+        var alphaPackage = new Package(
+            new PackageURL("pkg:maven/org.example/package@1"),
+            new DateTimeOffset(2021, 12, 13, 14, 15, 16, TimeSpan.FromHours(-4)));
+        var betaPackage = new Package(
+            new PackageURL("pkg:maven/org.example/package@2"),
+            alphaPackage.ReleasedAt.AddMonths(1));
```
edwinkortman, line 30: Oeh, I like how you wrote this.

edwinkortman (approved):
very nice, I'm surprised we aren't getting much better results from caching this. I wonder if something else could be slowing us down.

Issue #359, opened 2022-10-21
Get CI passing on the `main` branch


mscottford:
From pull requests that we decided to pull in early CI has gotten into a failing state. This needs to be resolved so that we can trust the tests again.

PR #372, opened 2022-10-25
Fix failing agents verify test

edwinkortman:
Related to #359 

There were a few problems with the AgentsVerifier that caused it not to run at all. I also took the opportunity to do a bit of refactoring:
- Use one loop for running DetectManifests
- Pass the targetDirectory, as we need an absolute URL for the agent. This was a repository URL before.
- The check for validating detected manifests versus processed manifests was always reporting false because the processed manifests in the loop is always 1
- Detect manifests only ran for Gemfiles


mscottford (approved):
This looks good to me. Thanks!

PR #378, opened 2022-10-26
Fix failing analyze feature tests

### LINKED ISSUE #359: Get CI passing on the `main` branch

edwinkortman:
Fixes #359
@mscottford I think this fix, and the fix introduced here https://github.com/corgibytes/freshli-cli/pull/372
Are cases where the suite is actually failing. If I run the tests locally, separately, they don't fail. If I run them all at once the first two of analyze.feature fail. Separating the realistic scenario into a different file only causes that one to fail. 

So I think for fixing the CI on the main branch we have to look at what you suggested in Slack to get GitHub actions running on a bigger computer.

### DIFF
--- a/features/analyze.feature
+++ b/features/analyze.feature
@@ -168,9 +168,8 @@ Feature: analyze
 
 
     Scenario: Run the analysis for a local directory
-        When I run `git clone https://github.com/corgibytes/freshli-fixture-java-test freshli-fixture-java-test`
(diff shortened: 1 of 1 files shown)


mscottford (approved):
This looks good to me. Nice and simple.

PR #381, opened 2022-10-30
Fix Build

mscottford:
The continuous integration builds have been failing for a while. This pull request fixes them.

* Increases command timeout
* Relaxes dependence on timezone when checking for correctness of date/time values
* Stops collecting code coverage information for the acceptance tests
* Fixes issue with the way the coverage results are sent to CodeClimate


mscottford:
The test suite takes about 16 minutes when run on my computer. On CI, it's taking about 1.5 hours.

PR #555, opened 2023-06-02
Increases test timeouts

mscottford:
Tests are failing on CI because they are timing out. This increases the test timeout in an attempt to resolve that issue.


mscottford:
Closing in favor of #572.

PR #692, opened 2023-09-25
Experimenting with larger GitHub Action worker

mscottford:


mscottford:
Closing this without merging it in. While the test suite does run a _little_ bit faster, the improvement does not justify the cost inrcrease.

